PR #1, opened 2023-05-30
chore: improves event handling

DavidMinarsch:


mech_client/interact.py
```diff
@@ -45,6 +47,19 @@
...
 PRIVATE_KEY_FILE_PATH = "ethereum_private_key.txt"
+EVENT_SIGNATURE_REQUEST = (
+    "0x4bda649efe6b98b0f9c1d5e859c29e20910f45c66dabfe6fad4a4881f7faf9cc"
+)
+EVENT_SIGNATURE_DELIVER = (
+    "0x3ec84da2cdc1ce60c063642b69ff2e65f3b69787a2b90443457ba274e51e7c72"
+)
+WSS_ENDPOINT = os.getenv(
+    "WEBSOCKET_ENDPOINT",
+    "wss://gno.getblock.io/3042bb06-d18e-4ce0-8ccf-83eedd35532b/mainnet/",
```
0xArdi, line 58: this wasnt a public endpoint

mech_client/interact.py
```diff
@@ -138,20 +180,28 @@ def watch_for_events(
...
+        if event_signature == EVENT_SIGNATURE_REQUEST:
             rich_logs = contract_instance.events.Request().processReceipt(tx_receipt)
             request_id = rich_logs[0]["args"]["requestId"]
             print(f"Request on-chain with id: {request_id}")
-        if count == 1:
+        if event_signature == EVENT_SIGNATURE_DELIVER:
             rich_logs = contract_instance.events.Deliver().processReceipt(tx_receipt)
             data = rich_logs[0]["args"]["data"]
+            request_id_ = rich_logs[0]["args"]["requestId"]
+            if request_id != request_id_:
```
0xArdi, line 192: so the assumption here is that there will always be a req before a deliver. What would happen if there's a deliver, request_id will be references before declared no?

mech_client/interact.py
DavidMinarsch: Ugh .. I copied the wrong one.

mech_client/interact.py
DavidMinarsch: good point - fixed now

PR #4, opened 2023-06-08
Dynamic tool loading

angrybayblade:
This PR adds support for loading tools dynamically from IPFS registry and define the agent dynamically. Please read the updated `README.md` for more information


mech_client/cli.py
```diff
@@ -1,37 +1,73 @@
...
+        interact(prompt=prompt, agent_id=agent_id, private_key_path=key, tool=tool)
+    except (ValueError, FileNotFoundError) as e:
+        raise click.ClickException(str(e)) from e
 
 
 @click.command()
 @click.argument("prompt")
 @click.argument("tool")
-def prompt_to_ipfs(prompt: str, tool: str) -> None:
+def _prompt_to_ipfs(prompt: str, tool: str) -> None:
```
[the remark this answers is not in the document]
DavidMinarsch: Same for the others

mech_client/cli.py
```diff
@@ -1,37 +1,73 @@
...
 
-@click.group(name="mechx")  # type: ignore
+@click.group(name="mechx")
 @click.version_option(__version__, prog_name="mechx")
 def cli() -> None:
     """Command-line tool for interacting with mechs."""
 
 
-@click.command()
+@click.command(name="interact")
```
DavidMinarsch, line 39: See below

mech_client/interact.py
```diff
@@ -178,39 +221,64 @@ def watch_for_events(
...
-    ethereum_crypto = EthereumCrypto(private_key_path=PRIVATE_KEY_FILE_PATH)
-    ethereum_ledger_api = EthereumApi(**ETHEREUM_TESTNET_CONFIG)
-    wss = register_event_handlers(ethereum_crypto)
-    contract_instance = send_request(ethereum_crypto, ethereum_ledger_api, prompt, tool)
-    response = watch_for_events(
-        wss, contract_instance, ethereum_ledger_api, ethereum_crypto
+def interact(
+    prompt: str,
+    agent_id: int,
+    tool: Optional[str] = None,
```
DavidMinarsch, line 253: tool is not optional?

also keep the order so that we don't unnecessarily make it harder to upgrade

angrybayblade: Tools IS optional here, if the user already know which tool they want to use they can provide it using `--tool` flag on cli and if they don't provide the tool name they'll be prompted with available tools to select from. Please read the updated `README.md`

mech_client/subgraph.py
```diff
@@ -0,0 +1,52 @@
...
+    """{
+    createMeches(where:{agentId:$agent_id}) {
+        mech
+    }
+}
+"""
+)
+
+
+def query_agent_address(agent_id: int) -> Optional[str]:
```
angrybayblade, line 39: To query the mech address using on chain agent ID

mech_client/interact.py
```diff
@@ -56,117 +64,152 @@
...
+    abi = json.loads(response["result"])
+    return ledger_api.get_contract_instance(
+        {"abi": abi, "bytecode": "0x"}, contract_address
+    )
+
+
+def _tool_selector_prompt(available_tools: List[str]) -> str:
+    """Tool selector prompt."""
+
+    tool_col_len = len(max(available_tools))
```
Adamantios, line 86:
```python
len(max(["c", "ab"])) == len("c") == 1
```

Would that be the desired column length or would it be `len("ab") == 2`?

mech_client/interact.py
```diff
@@ -56,117 +64,152 @@
...
+    id_col_len = max(2, len(str(len(available_tools))))
```
Adamantios, line 87: Why max 2?

mech_client/interact.py
angrybayblade: Fixed [43dd037](https://github.com/valory-xyz/mech-client/pull/4/commits/43dd0377478a109a0a0c27d30a27174a75ff8e1c)

Adamantios (approved):

PR #10, opened 2023-06-25
House keeping

angrybayblade:
- Add workflow
- Bump open-aea and open-autonomy
- Linters clean up

### DIFF
--- a/mech_client/acn.py
+++ b/mech_client/acn.py
@@ -69,7 +69,15 @@ CERT_REQUESTS = [
 
 
 def issue_certificate(cert_request: CertRequest, crypto: Crypto) -> None:
-    """Issue ACN certificate."""
+    """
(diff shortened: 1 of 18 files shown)


mech_client/acn.py
```diff
@@ -66,27 +67,41 @@
     }
 ]
 
-IPFS_REQUEST_URL = "https://gateway.autonolas.tech/ipfs/{hash_str}/{request_id}"
```
DavidMinarsch, line 69: Where has this gone?

DavidMinarsch (approved):

mech_client/acn.py
angrybayblade: This was unused, the URL is directly being formatted here https://github.com/valory-xyz/mech-client/blob/main/mech_client/acn.py#L138

PR #14, opened 2023-07-26
chore: bump open-autonomy and open-aea

0xArdi:
This PR bumps the mech-client to latest framework versions.

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -1,8 +1,8 @@
 {
     "dev": {},
     "third_party": {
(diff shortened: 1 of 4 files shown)


Adamantios (commented):
You will also need to update the versions in case you need to release:

1. https://github.com/valory-xyz/mech-client/blob/e25ddfcf99b41bb62aa2bf3f791485f5159f40dd/mech_client/__init__.py#L3
2. https://github.com/valory-xyz/mech-client/blob/e25ddfcf99b41bb62aa2bf3f791485f5159f40dd/pyproject.toml#L3

which are mismatching btw.

Adamantios (approved):

angrybayblade (approved):

PR #25, opened 2023-12-05
Fix deliver signature

angrybayblade:


mech_client/wss.py
```diff
@@ -35,7 +35,7 @@
     "0x4bda649efe6b98b0f9c1d5e859c29e20910f45c66dabfe6fad4a4881f7faf9cc"
 )
 EVENT_SIGNATURE_DELIVER = (
-    "0x3ec84da2cdc1ce60c063642b69ff2e65f3b69787a2b90443457ba274e51e7c72"
+    "0x0cd979445339c62199996f208428d987b1cea24d18e62b79ec24d94b636e8b70"
```
Adamantios, line 38: This depends on the agent id that the user gives. Therefore, we could use a mapping:

```python
EVENT_SIGNATURE_DELIVER = {
    ?: "0x3ec84da2cdc1ce60c063642b69ff2e65f3b69787a2b90443457ba274e51e7c72",
    3: "0x0cd979445339c62199996f208428d987b1cea24d18e62b79ec24d94b636e8b70",
}
```

PR #41, opened 2024-04-18
Relax dependencies versions

kongzii:
Similarly, as in https://github.com/valory-xyz/open-aea/pull/736, relax dependencies so it's easier to use this in other projects.


angrybayblade (approved):

kongzii:
@angrybayblade Thanks for merge here as well, is it possible to do release on both of these libraries, so we can use them, please?

angrybayblade:
@kongzii We'll try to get these releases out asap

PR #43, opened 2024-04-26
Bump framework version

angrybayblade:


mech_client/__init__.py
```diff
@@ -1,3 +1,3 @@
 """Mech client."""
 
-__version__ = "0.2.12"
```
angrybayblade, line 3: This was left out on https://github.com/valory-xyz/mech-client/pull/41

0xArdi (approved):

Adamantios (approved):

PR #76, opened 2025-03-14
Feat/nvm token

KahanMajmudar:


mech_client/marketplace_interact.py
```diff
@@ -726,14 +738,16 @@ def marketplace_interact(  # pylint: disable=too-many-arguments, too-many-locals
     mech_marketplace_request_config.delivery_rate = max_delivery_rate
     mech_marketplace_request_config.payment_type = payment_type
 
+    with open(Path(__file__).parent / "abis" / "IMech.json", encoding="utf-8") as f:
```
Adamantios, line 741: Please set a constant here, too.

Adamantios (changes requested):

mech_client/marketplace_interact.py
Adamantios, line 1: Let's tidy this up:

1. Create an enum for the payment types (native, token, etc.).
2. Create a dataclass for the payments:
    ```python
    @dataclass(frozen=True)
    class Payment:
        name: str
        address: str
        abi_path: Path | None = None
    ```
3. Create and use a dictionary of `Payments` as a constant, with `PaymentType` as keys and `Payment` as values.

Adamantios (changes requested):

mech_client/marketplace_interact.py
```diff
@@ -586,51 +588,33 @@ def check_prepaid_balances(
...
-            print("  - Please use scripts/deposit_native.py to add balance")
-            sys.exit(1)
 
-    if payment_type == PAYMENT_TYPE_TOKEN:
-        with open(
-            Path(__file__).parent / "abis" / "BalanceTrackerFixedPriceToken.json",
-            encoding="utf-8",
-        ) as f:
-            abi = json.load(f)
+    payment_type_name = (PaymentType(payment_type).name).lower()
```
Adamantios: Is there any case for the `payment_type` to be unsupported? If yes, then we need to make this code safer as `PaymentType(payment_type)` would raise.

mech_client/marketplace_interact.py
```diff
@@ -590,7 +590,7 @@ def check_prepaid_balances(
     requester = crypto.address
 
     payment_type_name = (PaymentType(payment_type).name).lower()
-    payment_type_abi_path = PAYMENT_TYPE_TO_ABI_PATH.get(payment_type)
+    payment_type_abi_path = cast(Path, PAYMENT_TYPE_TO_ABI_PATH.get(payment_type))
```
Adamantios: You can specify the type of the dict during definition instead.

mech_client/marketplace_interact.py
```diff
@@ -568,51 +588,33 @@ def check_prepaid_balances(
...
+        contract_address=mech_payment_balance_tracker,
+        abi=abi,
+        ledger_api=ledger_api,
+    )
+    requester_balance = balance_tracker_contract.functions.mapRequesterBalances(
+        requester
+    ).call()
+    if requester_balance < max_delivery_rate:
+        print(
+            f"  - Sender {payment_type_name.lower()} deposited balance low. Needed: {max_delivery_rate}, Actual: {requester_balance}"
```
Adamantios:
```suggestion
            f"  - Sender {payment_type_name} deposited balance low. Needed: {max_delivery_rate}, Actual: {requester_balance}"
```

mech_client/marketplace_interact.py
```diff
@@ -568,51 +588,33 @@ def check_prepaid_balances(
...
-        if requester_balance < max_delivery_rate:
-            print(
-                f"  - Sender Token deposited balance low. Needed: {max_delivery_rate}, Actual: {requester_balance}"
-            )
-            print(f"  - Sender Address: {requester}")
-            print("  - Please use scripts/deposit_token.py to add balance")
-            sys.exit(1)
+        print(f"  - Sender Address: {requester}")
+        print(
+            f"  - Please use scripts/deposit_{payment_type_name.lower()}.py to add balance"
```
Adamantios:
```suggestion
            f"  - Please use scripts/deposit_{payment_type_name}.py to add balance"
```

mech_client/marketplace_interact.py
```diff
@@ -726,22 +728,24 @@ def marketplace_interact(  # pylint: disable=too-many-arguments, too-many-locals
     mech_marketplace_request_config.delivery_rate = max_delivery_rate
     mech_marketplace_request_config.payment_type = payment_type
 
+    with open(Path(__file__).parent / "abis" / "IMech.json", encoding="utf-8") as f:
```
Adamantios, line 732:
```suggestion
    with open(IMECH_ABI_PATH, encoding="utf-8") as f:
```

mech_client/marketplace_interact.py
```diff
@@ -220,11 +238,13 @@ def fetch_requester_nvm_subscription_balance(
...
     :param mech_payment_balance_tracker: Requested mech's balance tracker contract address
     :type mech_payment_balance_tracker: str
+    :param payment_type: Requested mech's payment type
+    :type payment_type: str
     :return: The requester balance.
     :rtype: int
     """
     with open(
-        Path(__file__).parent / "abis" / "BalanceTrackerNvmSubscriptionNative.json",
+        PAYMENT_TYPE_TO_ABI_PATH[payment_type],
```
Adamantios, line 247: Access the dict safely.

mech_client/marketplace_interact.py
```diff
@@ -772,11 +776,14 @@ def marketplace_interact(  # pylint: disable=too-many-arguments, too-many-locals
             max_delivery_rate,
         )
 
-    if payment_type == PAYMENT_TYPE_NVM:
-        print("Nevermined Mech detected, subscription credits to be used")
+    if payment_type in [PaymentType.NATIVE_NVM.value, PaymentType.TOKEN_NVM.value]:
+        nvm_mech_type = (PaymentType(payment_type).name).lower()
```
Adamantios, line 781:
```suggestion
        nvm_mech_type = PaymentType(payment_type).name.lower()
```

Adamantios (changes requested):

mech_client/marketplace_interact.py
KahanMajmudar: Should be safe as I am checking for payment types here: https://github.com/valory-xyz/mech-client/blob/feat/nvm-token/mech_client/marketplace_interact.py#L780

Adamantios: You mean here right?

https://github.com/valory-xyz/mech-client/blob/eb3f136b76f9599249a380226c432a3d8b87690b/mech_client/marketplace_interact.py#L775

Yeap, makes sense.

mech_client/marketplace_interact.py
KahanMajmudar: Should not be an issue as I do the check inside the fetch mech info

KahanMajmudar: https://github.com/valory-xyz/mech-client/blob/feat/nvm-token/mech_client/marketplace_interact.py#L160

Adamantios: ok, keep in mind that it would only be safe if the method accepted a `PaymentType`, which could be obtained after the check you linked because the code might be restructured in the future.

Adamantios (approved):

PR #82, opened 2025-04-09
Misc/fix

KahanMajmudar:


mech_client/marketplace_interact.py
```diff
@@ -108,14 +108,10 @@ class PaymentType(Enum):
 
 CHAIN_TO_DEFAULT_MECH_MARKETPLACE_REQUEST_CONFIG = {
     100: {
-        "mech_marketplace_contract": "0x735FAAb1c4Ec41128c367AFb5c3baC73509f70bB",
-        "priority_mech_address": "0x478ad20eD958dCC5AD4ABa6F4E4cc51e07a840E4",
         "response_timeout": 300,
         "payment_data": "0x",
     },
     8453: {
-        "mech_marketplace_contract": "0xf24eE42edA0fc9b33B7D41B06Ee8ccD2Ef7C5020",
```
mariapiamo, line 117: we should keep the MM contract

mech_client/marketplace_interact.py
```diff
@@ -108,14 +108,10 @@ class PaymentType(Enum):
 
 CHAIN_TO_DEFAULT_MECH_MARKETPLACE_REQUEST_CONFIG = {
     100: {
-        "mech_marketplace_contract": "0x735FAAb1c4Ec41128c367AFb5c3baC73509f70bB",
```
mariapiamo, line 111: we should keep the MM contract

mariapiamo (changes requested):

mech_client/marketplace_interact.py
KahanMajmudar: I am taking the info from the mech configs, so removed it here to not have any mismatch in the future as well

mariapiamo (approved):

Adamantios (approved):

PR #89, opened 2025-05-22
Test/nvm sub contract

KahanMajmudar:


mariapiamo (approved):

mariapiamo (changes requested):
While testing, I observed the following:
PROVIDERS: ['0x0000000000000000000000000000000000000000', '0x0000000000000000000000000000000000000000', '0x0000000000000000000000000000000000000000', '0x4a2f40E14309c20c0C3803c3CcCd5E9B5F2D4eCA']

First 3 providers are removed, maybe this is the reason we see those zero. I think we need to fix this

PR #96, opened 2025-07-04
fix: misc issues faced while using mech client using pip install

KahanMajmudar:


scripts/nvm_subscription/contracts/base_contract.py
```diff
@@ -41,7 +41,9 @@ def _load_contract_info(self) -> Dict[str, Any]:
         Returns:
             dict: A dictionary containing the contract address and ABI.
         """
-        path = os.path.join('mech_client', 'abis', f'{self.name}.{self.chain_name}.json')
+        current_dir = os.path.dirname(os.path.abspath(__file__))
+        root_dir = os.path.abspath(os.path.join(current_dir, '..', '..', '..'))
```
Totoual, line 45: Why don't we use os.getcwd() and then navigate from there ?

KahanMajmudar: `__file__` is relative to the location of this particular script and so even if it's run from different working dir, it's more reliable

Totoual (approved):

mariapiamo (approved):

PR #116, opened 2025-09-17
WIP: integrate operate

KahanMajmudar:


config/mech_client.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "Mech Client (agent mode)",
+    "hash": "bafybeigb5a6ptwtfx5wpnvqv6mp5jg6dpjt5scgjgnq74d4mri2i7vntbm",
```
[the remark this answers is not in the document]
KahanMajmudar: Yes, as per the spec doc, we still have to figure out a dummy agent and service for mech client

config/mech_client.json
```diff
@@ -0,0 +1,25 @@
...
+    "description": "The mech client requests tasks to AI mechs.",
+    "image": "https://gateway.autonolas.tech/ipfs/bafybeidzpenez565d7vp7jexfrwisa2wijzx6vwcffli57buznyyqkrceq",
+    "service_version": "v0.9.0",
+    "home_chain": "gnosis",
+    "configurations": {
+        "gnosis": {
+            "agent_id": 37,
```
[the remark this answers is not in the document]
mariapiamo: To properly associate the services with the real usage via events, ie mech-client agent, it is necessary to to mint a new agent on L1. 
Before doing that, we need: 
- A Package hash (e.g. https://gateway.autonolas.tech/ipfs/bafybeihae4lqokbjzcu5reb4w7cf5xy6fuibkejjqw24chdraqqlpbvpwu)
- a basic component (e.g. https://marketplace.olas.network/ethereum/components/134)  
Please check also with Jose.

For the time being, can we avoid to use mech agent id? Please consider to temporely use agent id = 1.

config/mech_client.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "Mech Client (agent mode)",
+    "hash": "bafybeicrsrqajrzgweh7q4grfsbkkadovyqzfouqgb5lmxyqulqv7mjaye",
+    "description": "The mech_client service to allow interacting with on-chain mechs",
+    "image": "https://gateway.autonolas.tech/ipfs/bafybeidzpenez565d7vp7jexfrwisa2wijzx6vwcffli57buznyyqkrceq",
+    "service_version": "v0.1.0",
+    "home_chain": "gnosis",
+    "configurations": {
+        "gnosis": {
```
jmoreira-valory, line 9: We don't have a configuration for the other chain?

mech_client/cli.py
```diff
@@ -58,10 +69,131 @@
...
+        }
+
+    if config.user_provided_args is None:
+        config.user_provided_args = {}
+
+    config.store()
+    return config
+
+
+def fetch_agent_mode_data(chain_config: Optional[str]) -> Tuple[str, str]:
```
jmoreira-valory, line 117: Why not use service_manager and keys_manager here instead of managing paths, etc. by hand ?

README.md
```diff
@@ -114,6 +114,26 @@ Once you have your API key, you'll need to configure it in your environment. Use
 export MECHX_API_KEY=<your api key>
 ```
 
+### Set up agent mode for on-chain interactions (For Latest Marketplace Mechs only)
```
mariapiamo: I would say that we need to re-org this readme.

My suggestion is:
let's have all the MM from top to botton with agent mode (without specifying that it is agent mode but providing the context on what will happen). Then let's have a subsection of the MM one with private key set, then we have a final section with legacy description.

mariapiamo: additionally, we need to the following mention where funds needs to be transferred in case of agent mode / non agent mode for the various payment models

scripts/nvm_subscription/manager.py
```diff
@@ -321,3 +325,30 @@ def create_subscription(self, did: str, wallet_pvt_key: str, chain_id: int) -> D
...
             return {"status": "failed", "receipt": dict(receipt)}
+
+
+    def check_sufficient_sender_balance(self) -> Tuple[bool, str]:
+        w3 = self.web3
+        sender = self.sender
+        chain_id = w3.eth.chain_id
+
+        # For gnosis, native xdai is used for purchase 
+        if chain_id == 100:
```
Totoual: Can we make it a const, like a dictionary or something ? So we don't have magic numbers in the code ? Same for base.

PR #164, opened 2026-02-06
feat: implement complete offchain mech support

DavidMinarsch:
Add full offchain request functionality that was previously missing. When use_offchain=True, the code now:
- Sends HTTP POST to offchain mech endpoint with signed request data
- Polls offchain endpoint for responses (not on-chain)
- Skips on-chain transaction entirely

Changes:
- Created OffchainDeliveryWatcher in domain/delivery/offchain_watcher.py
  - Polls {mech_offchain_url}/fetch_offchain_info for responses
  - Async polling with configurable timeout
- Updated MarketplaceService.send_request() to branch between on-chain/off-chain flows
- Added _send_offchain_request() method:
  - POSTs to {mech_offchain_url}/send_signed_requests with:
    - sender, signature, ipfs_hash, request_id, delivery_rate, nonce, ipfs_data
  - Calculates request_id via marketplace contract
  - Signs request_id with private key
- Added fetch_ipfs_hash() in infrastructure/ipfs/metadata.py
  - Creates metadata and computes IPFS hash without uploading
  - Returns hash + JSON data for offchain transmission
- Updated domain/delivery/__init__.py to export OffchainDeliveryWatcher

Previously, use_offchain only skipped the on-chain delivery watcher but still sent an on-chain transaction. Now it properly handles the complete offchain flow matching historic implementation.

### DIFF
--- a/tests/unit/domain/test_delivery_watchers.py
+++ b/tests/unit/domain/test_delivery_watchers.py
@@ -83,64 +83,30 @@ class TestOnchainDeliveryWatcherWatch:
     """Tests for OnchainDeliveryWatcher watch method."""
 
     @pytest.mark.anyio
-    @patch("mech_client.domain.delivery.onchain_watcher.time.sleep")
-    @patch("mech_client.domain.delivery.onchain_watcher.get_abi")
     async def test_watch_single_request_immediate_delivery(
-        self, mock_get_abi: MagicMock, mock_sleep: MagicMock, mock_web3_contract: MagicMock, mock_ledger_api: MagicMock
+        self, mock_web3_contract: MagicMock, mock_ledger_api: MagicMock
     ) -> None:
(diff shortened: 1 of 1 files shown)


mech_client/infrastructure/ipfs/metadata.py
```diff
@@ -28,6 +28,49 @@
 from mech_client.infrastructure.ipfs.client import IPFSClient
 
 
+def fetch_ipfs_hash(
```
DIvyaNautiyal07: The function is meant to calculate ipfs hash without uploading (for off-chain requests), but the code calls client.upload(file_name) which runs ipfs_tool.client.add(file_path, pin=True), this pins the file to IPFS.

DavidMinarsch: https://github.com/valory-xyz/mech-client/pull/166

PR #234, opened 2026-05-18
chore: bump open-autonomy 0.21.22, open-aea 2.2.6

DIvyaNautiyal07:
## Summary

Bumps mech-client onto the same OA / open-aea version the rest of the fleet just moved to.

- `[project] dependencies`: `open-aea-helpers==0.21.20 → 0.21.22` (the only direct framework pin in this repo)
- `[tool.tomte]` metadata: `open_autonomy_version "0.21.20" → "0.21.22"` and `open_aea_version "2.2.3" → "2.2.6"` (drives the version pins tomte injects into `liccheck` / `check-third-party-hashes`)
- `tox.ini`: `[tomte-extensions] extra_deps` `open-autonomy[all]==0.21.20 → 0.21.22` and `[Authorized Packages] open-autonomy: ==0.21.20 → 0.21.22`
- `packages/valory/agents/mech_client/aea-config.yaml`: stale dev plugin pin `open-aea-ledger-ethereum: ==1.65.0 → ==2.2.2` (see note 2 below)
- `packages/valory/services/mech_client/service.yaml` + `packages/packages.json`: dev fingerprints rehashed by `autonomy packages lock` (agent CID changed, service references it by CID)
- `uv.lock`: regenerated; only direct upgrade is `open-aea-helpers 0.21.20 → 0.21.22`

Non-breaking patch upgrades per upstream upgrading.md
([open-autonomy 0.21.21 → 0.21.22](https://github.com/valory-xyz/open-autonomy/blob/v0.21.22/docs/upgrading.md),
[open-aea 2.2.3 → 2.2.6](https://github.com/valory-xyz/open-aea/blob/v2.2.6/docs/upgrading.md)).
Upstream explicitly says "No agent / package / config edits are required" beyond pin changes.

## Notes specific to this repo

1. **mech-client doesn't directly pin `open-autonomy` or `open-aea` in `[project] dependencies`** — it only pins `open-aea-helpers`. The framework comes in transitively via `olas-operate-middleware` (currently resolves to `open-autonomy==0.21.19` / `open-aea==2.2.1` / `open-aea-ledger-ethereum==2.2.2` in the lockfile, bounded by `olas-operate-middleware<0.16`). The test/lint envs install `open-autonomy[all]==0.21.22` directly via `[tomte-extensions] extra_deps`. If the goal is for SDK consumers to also get 0.21.22 / 2.2.6 at install time, that's a separate follow-up (bump `olas-operate-middleware` or add a direct framework pin to `[project] dependencies`).

2. **Stale `==1.65.0` plugin pin in `aea-config.yaml`** has been bumped to `==2.2.6` (matches the OA / open-aea bump target). The pin is inert metadata in this repo — no CI job runs `aea install`, `autonomy fetch`, `build-image`, or `autonomy packages lock --check` against it, so the value carries no install-time semantics. Two alignment options were discussed during review: align with the **resolved** framework in `uv.lock` (`==2.2.2`, transitively from `olas-operate-middleware<0.16`), or align with the **target** framework on this PR (`==2.2.6`, what every other dev YAML in the fleet pins to). Settled on the target to keep the value consistent with the rest of the fleet and with the PR's own framing — the underlying resolved-vs-target divergence is the note-1 follow-up (bump `olas-operate-middleware` or add a direct framework pin in `[project] dependencies`). Last touched 2025-10-02 (commit `3c2defba`); the Oct 2025 → Wave-2a → 0.21.16rc1 → 0.21.19 → 0.21.20 bumps all missed this site, which is why the value had drifted so far.

3. **No docker image pin** in `release.yml` / `workflow.yml` — mech-client is a PyPI-published SDK, not an agent service, so there's no `valory/open-autonomy-user:<tag>` to bump.

## Test plan

Matches `.github/workflows/workflow.yml` exactly.

- [x] `uv lock --check`
- [x] `uv sync --all-groups`
- [x] `tomte check-copyright --author valory` (pre-existing local-only failure on 3 gitignored empty `__init__.py` files; these are not in the repo, won't reach CI)
- [x] `tomte tox -e liccheck`
- [x] `tomte tox -e check-third-party-hashes`
- [x] `tomte tox -p -e safety -e bandit`
- [x] `tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint`
- [x] `tomte tox -e gitleaks`
- [x] `uv run pytest tests/unit -v --tb=short -k "not trio" --cov=mech_client --cov-fail-under=100` — **702 passed, 100.00% coverage maintained**
- [x] `autonomy packages lock --check`
- [x] `aea --version` reports `2.2.1` (transitive via olas-operate-middleware), test-env `open-autonomy==0.21.22` installs cleanly

### DIFF
--- a/packages/valory/agents/mech_client/aea-config.yaml
+++ b/packages/valory/agents/mech_client/aea-config.yaml
@@ -46,6 +46,6 @@ logging_config:
       propagate: true
 dependencies:
   open-aea-ledger-ethereum:
-    version: ==1.65.0
+    version: ==2.2.6
 customs: []
 default_connection: null
--- a/packages/valory/agents/mech_client/aea-config.yaml
+++ b/packages/valory/agents/mech_client/aea-config.yaml
@@ -46,6 +46,6 @@ logging_config:
       propagate: true
 dependencies:
   open-aea-ledger-ethereum:
-    version: ==2.2.6
+    version: ==2.2.2
 customs: []
 default_connection: null
--- a/packages/valory/agents/mech_client/aea-config.yaml
+++ b/packages/valory/agents/mech_client/aea-config.yaml
@@ -46,6 +46,6 @@ logging_config:
       propagate: true
 dependencies:
   open-aea-ledger-ethereum:
-    version: ==2.2.2
+    version: ==2.2.6
 customs: []
 default_connection: null
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -1,7 +1,7 @@
 {
     "dev": {
-        "agent/valory/mech_client/0.1.0": "bafybeiaudemcmceh4iukdhezxxhdypu2hwcrvrkiwl5zvhn34le2ucmziu",
-        "service/valory/mech_client/0.1.0": "bafybeihlke2upscxnrkq2gae4dsxzzncwpdqxhfawvw3zn6sesgjmwcft4"
+        "agent/valory/mech_client/0.1.0": "bafybeif3azn4azaopsup5lrh4r2ixvkd7in2zgwhlqckut5qphsfgrfnc4",
+        "service/valory/mech_client/0.1.0": "bafybeie2muu4byu34wxb5hbc2jw7zaqunzqpfx6cm2wiycmezvtvdvei2a"
     },
     "third_party": {
         "protocol/valory/abci/0.1.0": "bafybeiak4pwac3pwjtd7weskxdhkrcopckr32vtazfbuteu6yqjkwarpte",
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -1,7 +1,7 @@
 {
     "dev": {
-        "agent/valory/mech_client/0.1.0": "bafybeif3azn4azaopsup5lrh4r2ixvkd7in2zgwhlqckut5qphsfgrfnc4",
-        "service/valory/mech_client/0.1.0": "bafybeie2muu4byu34wxb5hbc2jw7zaqunzqpfx6cm2wiycmezvtvdvei2a"
+        "agent/valory/mech_client/0.1.0": "bafybeianp75zchjncfaor57f3ivwvdgwoihxfydl7uwlvh22mhqgjs2rgq",
+        "service/valory/mech_client/0.1.0": "bafybeibvpr5b6ccwuflzqhll2hczcpcjixbobabj5hoj5nx2s6azsqggxy"
     },
     "third_party": {
         "protocol/valory/abci/0.1.0": "bafybeiak4pwac3pwjtd7weskxdhkrcopckr32vtazfbuteu6yqjkwarpte",
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -1,7 +1,7 @@
 {
     "dev": {
-        "agent/valory/mech_client/0.1.0": "bafybeianp75zchjncfaor57f3ivwvdgwoihxfydl7uwlvh22mhqgjs2rgq",
-        "service/valory/mech_client/0.1.0": "bafybeibvpr5b6ccwuflzqhll2hczcpcjixbobabj5hoj5nx2s6azsqggxy"
+        "agent/valory/mech_client/0.1.0": "bafybeif3azn4azaopsup5lrh4r2ixvkd7in2zgwhlqckut5qphsfgrfnc4",
+        "service/valory/mech_client/0.1.0": "bafybeie2muu4byu34wxb5hbc2jw7zaqunzqpfx6cm2wiycmezvtvdvei2a"
     },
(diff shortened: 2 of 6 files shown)


DhairyaPatel7 (commented):
Clean, mechanical patch-version bump (open-aea-helpers 0.21.20→0.21.22, tomte open_autonomy/open_aea metadata, tox lint pins, agent plugin pin, regenerated CIDs). All bumps are consistent, no stale references remain, and the PR description preempts the obvious concerns (transitive open-autonomy still 0.21.19 via olas-operate-middleware<0.16, large 1.65.0→2.2.6 ledger-plugin jump is a stale-pin cleanup under aea_version>=2.0.0). No findings.

packages/valory/agents/mech_client/aea-config.yaml
```diff
@@ -46,6 +46,6 @@ logging_config:
       propagate: true
 dependencies:
   open-aea-ledger-ethereum:
-    version: ==1.65.0
+    version: ==2.2.6
```
LOCKhart07: The resolved framework on this branch is `open-aea==2.2.1` / `open-aea-ledger-ethereum==2.2.2` / `open-autonomy==0.21.19` (`uv.lock`, transitive via `olas-operate-middleware<0.16`). `==2.2.6` matches neither — it only matches the tomte lint-env metadata target, which `pyproject.toml` itself scopes to `liccheck`/`check-third-party-hashes` pin rendering, not installed deps.

Non-blocking: no CI job does `aea install` / `autonomy fetch` / `build-image`, and there's no `autonomy packages lock --check`, so this pin is inert metadata in a never-installed dev package. But the PR frames it as a correctness fix, and "matching the OAEA bump" (note 2) is misleading — the old stale `1.65.0` and the new `2.2.6` are both decoupled from reality, just in different directions.

Recommendation: either pin to the resolved `==2.2.2` (or `==2.2.1` to track core), or reword note 2 to state plainly that this pin is intentionally decoupled from the transitively-resolved framework and set to the tomte target as dev-package metadata only — so the next bumper isn't confused about which version world it lives in.

DhairyaPatel7 (dismissed):
Clean bump. Pins move consistently across pyproject.toml / tox.ini / tomte metadata, the stale `==1.65.0` plugin pin in `packages/valory/agents/mech_client/aea-config.yaml` is brought current to `==2.2.6`, and the cascaded agent/service CIDs match what `autonomy packages lock` should produce. CI green across the matrix. Description is thorough — the note that mech-client only pins `open-aea-helpers` directly (framework comes in transitively via `olas-operate-middleware<0.16`) is the right call-out.

Non-blocking nit: `==1.65.0 → ==2.2.6` is a major-version delta in the dev `aea-config.yaml`. Worth a quick smoke at `aea install` / `aea run` against the regenerated agent CID before relying on it locally, just to confirm the v2 plugin loads cleanly under the current agent skill set.

packages/valory/agents/mech_client/aea-config.yaml
DIvyaNautiyal07: You're right, the framing was muddled. The yaml pin and the tomte target were sitting in two different version worlds and note 2 hand-waved over that.

Pinned it to `==2.2.2` to match the resolved `open-aea-ledger-ethereum` in `uv.lock` (transitive via `olas-operate-middleware<0.16`) rather than the tomte lint-env target. The yaml is inert metadata in this repo so the value doesn't change install behaviour either way, but aligning it to what's actually installed gives the next bumper one source of truth (uv.lock) instead of two.

Updated note 2 in the PR description to call this out explicitly. `autonomy packages lock` cascaded the fingerprint change into the agent and service CIDs (4 lines, 3 files).

packages/valory/agents/mech_client/aea-config.yaml
```diff
@@ -46,6 +46,6 @@ logging_config:
       propagate: true
 dependencies:
   open-aea-ledger-ethereum:
-    version: ==1.65.0
+    version: ==2.2.2
```
Adamantios:
```suggestion
    version: ==2.2.6
```

DhairyaPatel7 (dismissed):

packages/valory/agents/mech_client/aea-config.yaml
DIvyaNautiyal07: Applied in ae0fe81 — flipped back to `==2.2.6` and re-locked the agent + service CIDs. Updated note 2 in the PR description to call out the resolved-vs-target divergence as the note-1 follow-up rather than something this PR resolves.

Adamantios (approved):

DhairyaPatel7 (approved):

PR #240, opened 2026-06-11
fix: client-mode USDC payment reliability

LOCKhart07:
## Summary

Three issues surface end-to-end when sending a paid USDC request in client mode against Polygon USDC. This PR fixes the two that live in mech-client and bumps the dependency that fixes the other one upstream. One small consistency cleanup is included alongside.

### 1. Bump `open-autonomy 0.21.23 → 0.21.25` (pulls in `open-aea 2.2.9`)

`EthereumApi.build_transaction` did not propagate `from` into the `tx_params` it built, so `update_with_gas_estimate` called `eth_estimateGas` with no `from`. The RPC therefore simulated as `msg.sender = 0x0`, and ERC20 / marketplace contracts that branch on `msg.sender` reverted at the estimation step:

- `approve` hitting `require(owner != address(0))` → `"ERC20: approve from the zero address"`
- balance-tracker `transferFrom` failing on `allowance[0x0]` → `"ERC20: transfer amount exceeds allowance"`

even though the real signed transaction would succeed.

Fixed in [valory-xyz/open-aea#925](https://github.com/valory-xyz/open-aea/pull/925), bundled in [valory-xyz/open-autonomy#2535](https://github.com/valory-xyz/open-autonomy/pull/2535) (released as `0.21.25`).

Pulling that in lets us drop the two `_is_gas_estimation_enabled = False` try/finally workarounds that earlier drafts of this PR carried in `TokenPaymentStrategy.approve_if_needed` and `MarketplaceService.send_request`.

### 2. `replacement transaction underpriced` (nonce race) — fixed

After `approve_if_needed`, `send_request` immediately built the request transaction on the pre-approval `latest` nonce while the approval was still pending in the mempool. Two failure modes:

- Both transactions grab the same nonce → "replacement transaction underpriced".
- Even when nonces differ, the request executes before the allowance is on-chain → "ERC20: transfer amount exceeds allowance".

**Fix** (`services/marketplace_service.py`): capture the approval tx hash, `wait_for_receipt` on it, and raise loudly with the tx hash if `status == 0`. The request is built only after the approval is mined and successful.

### 3. Approve reverts out of gas on Circle's USDC proxy (Polygon) — fixed

`approve_if_needed` hardcoded `gas=60000`. Circle's USDC proxy on Polygon (`0x3c499c542cef5e3811e1192ce70d8cc03d5c3359`) needs ~67k for a 0→nonzero approve due to delegatecall overhead and storage writes. Result: approve reverts out-of-gas (gasUsed 59655 / 60000 — classic EIP-150 1/64 retention signature with `status == 0`).

This was previously silent because the code never checked the approve receipt's status before sending the request — #2 above now also catches it loudly.

**Fix** (`domain/payment/token.py`): bump the gas fallback to 120000. With gas estimation now working correctly on `open-aea 2.2.9`, this only matters when estimation fails (RPC quirk, etc.), but it's the right safety net.

Repro tx (first v3 superforcaster attempt during investigation): https://polygonscan.com/tx/0xf70f7bc29ac42fbc1b81b99ea0786b054f3653bb116d7420e0291b34f7c0fdba

### Cleanup — approve amount sourced from the per-mech rate, not the chain default

`MarketplaceService.send_request` was computing the approve amount as `self.mech_config.price * len(prompts)`. `mech_config.price` is a *static chain-wide default* from `mech_client/configs/mechs.json` (e.g. `10000` for polygon). The marketplace actually pulls `deliveryRate * numRequests` via `BalanceTracker.checkAndRecordDeliveryRates`, where `deliveryRate` is read live from the priority mech's `maxDeliveryRate()`. Today every mech we tested charges the chain default, so the two numbers coincide and nothing breaks. The moment a mech ever sets a different rate, the approve would under-fund and the marketplace's `transferFrom` would revert with "ERC20: transfer amount exceeds allowance".

**Change** (`services/marketplace_service.py`): use the `max_delivery_rate` already fetched from the mech contract for the approve amount, matching what the marketplace will actually pull. No behaviour change for any current mech.

```python
- price = self.mech_config.price * len(prompts)
+ price = max_delivery_rate * len(prompts)
```

## End-to-end verification

Six successful client-mode Polygon USDC requests after the fixes (3 mechs × 2 tools):

| Service | Tool | Tx |
|---|---|---|
| 21 | superforcaster-polymarket-v1 | [0x5b6cdb…453c](https://polygonscan.com/tx/0x5b6cdb46069dfdbf2f92e2f4fb0c1764d0ac90f805936896103dcdc4746f453c) |
| 25 | superforcaster-polymarket-v1 | [0xb2e3f0…0504](https://polygonscan.com/tx/0xb2e3f0c9a950c65e7526e2d5454e42f56307f710c076195edfa47682200b0504) |
| 44 | superforcaster-polymarket-v1 | [0x5a8a8c…8c60](https://polygonscan.com/tx/0x5a8a8c178c77c570a8db9db1b9d903f0b8bad1f3f49061650d5a105213d48c60) |
| 21 | superforcaster-polymarket-v2 | [0x7b3379…3d61](https://polygonscan.com/tx/0x7b33796bc9ceb1578500e079f7eb9135764807124cf59f8347c984de94313d61) |
| 25 | superforcaster-polymarket-v2 | [0x6f3a02…1f22](https://polygonscan.com/tx/0x6f3a0271d74ae87a2cd1fd2805ce58f0a487fdc37655c4a6a44b9da7c3861f22) |
| 44 | superforcaster-polymarket-v2 | [0xec34de…b67c](https://polygonscan.com/tx/0xec34de24cf4faa016853d26df1b8ef8e96819e9bbd66fcef7b790a61ecfcb67c) |

Those verifications were done with the workaround-disable-gas-estimation form in place. The current PR removes those workarounds and relies on `open-aea 2.2.9`'s correct `eth_estimateGas` behaviour, which makes the same end-to-end path equivalent (now via working estimation, previously via the hardcoded gas fallback).

## Files changed

- `pyproject.toml`, `tox.ini`, `uv.lock` — `open-aea-helpers 0.21.23 → 0.21.25`; cascades to `open-aea 2.2.9` / `open-autonomy 0.21.25` / `open-aea-ledger-ethereum 2.2.9` / `open-aea-cli-ipfs 2.2.9`
- `mech_client/domain/payment/token.py` — gas bump + drop the open-aea zero-address workaround in `approve_if_needed`
- `mech_client/services/marketplace_service.py` — `wait_for_receipt` + status check after approve, per-mech approve amount, drop the open-aea zero-address workaround around request submission
- `tests/unit/domain/test_payment_strategies.py` — update `test_approve_if_needed_with_executor` gas assertion (60000 → ≥67200)

## Test plan

- [x] `tomte tox -e black-check,isort-check,flake8,mypy,pylint,bandit` — all green locally
- [x] `uv run pytest tests/unit -k "not trio"` — 726 passed
- [x] Manual client-mode mech request on Polygon (USDC) succeeds end-to-end (see table above)
- [ ] Re-run the manual Polygon USDC request on `0.21.25` + the workaround-removed code to confirm equivalence end-to-end via the corrected estimation path
- [ ] Manual client-mode mech request on Gnosis (xDAI native, untouched path) still succeeds

### DIFF
--- a/mech_client/domain/payment/token.py
+++ b/mech_client/domain/payment/token.py
@@ -114,17 +114,37 @@ class TokenPaymentStrategy(PaymentStrategy):
         abi = get_abi("IToken.json")
         token_contract = get_contract(token_address, abi, self.ledger_api)
 
-        # Build approval transaction arguments
-        tx_args = {"sender_address": payer_address, "value": 0, "gas": 60000}
+        # Build approval transaction arguments.
+        # Gas is a fixed fallback used when gas estimation is disabled; some
+        # tokens (e.g. Circle USDC proxy on Polygon) need ~67k for a 0->nonzero
+        # approve, so 60k underfunds it and the approve reverts out of gas.
+        tx_args = {"sender_address": payer_address, "value": 0, "gas": 120000}
         method_name = "approve"
         method_args = {"_to": spender_address, "_value": amount}
 
-        return executor.execute_transaction(
-            contract=token_contract,
-            method_name=method_name,
-            method_args=method_args,
-            tx_args=tx_args,
+        # Workaround for valory-xyz/open-aea#924: EthereumApi.build_transaction
+        # does not propagate `from` into the tx dict it builds, so gas
+        # estimation runs with msg.sender = 0x0 and ERC20 approve reverts with
+        # "approve from the zero address". Disable gas estimation locally so
+        # the hardcoded gas fallback above is used. Remove once the upstream
+        # fix lands.
+        original_gas_estimation = getattr(
+            self.ledger_api, "_is_gas_estimation_enabled", False
         )
+        try:
+            self.ledger_api._is_gas_estimation_enabled = (  # noqa: SLF001  # pylint: disable=protected-access
+                False
+            )
+            return executor.execute_transaction(
+                contract=token_contract,
+                method_name=method_name,
+                method_args=method_args,
+                tx_args=tx_args,
+            )
+        finally:
+            self.ledger_api._is_gas_estimation_enabled = (  # noqa: SLF001  # pylint: disable=protected-access
+                original_gas_estimation
+            )
 
     def get_balance_tracker_address(self) -> str:
         """
--- a/mech_client/domain/payment/token.py
+++ b/mech_client/domain/payment/token.py
@@ -84,7 +84,7 @@ class TokenPaymentStrategy(PaymentStrategy):
         balance = token_contract.functions.balanceOf(payer_address).call()
         return balance >= amount
 
-    def approve_if_needed(  # pylint: disable=too-many-locals
+    def approve_if_needed(
         self,
         payer_address: str,
         spender_address: str,
@@ -114,37 +114,19 @@ class TokenPaymentStrategy(PaymentStrategy):
         abi = get_abi("IToken.json")
         token_contract = get_contract(token_address, abi, self.ledger_api)
 
-        # Build approval transaction arguments.
-        # Gas is a fixed fallback used when gas estimation is disabled; some
-        # tokens (e.g. Circle USDC proxy on Polygon) need ~67k for a 0->nonzero
-        # approve, so 60k underfunds it and the approve reverts out of gas.
+        # Fallback gas if estimation fails. Some tokens (e.g. Circle USDC
+        # proxy on Polygon) need ~67k for a 0->nonzero approve; 60k
+        # underfunds it.
         tx_args = {"sender_address": payer_address, "value": 0, "gas": 120000}
         method_name = "approve"
         method_args = {"_to": spender_address, "_value": amount}
 
-        # Workaround for valory-xyz/open-aea#924: EthereumApi.build_transaction
-        # does not propagate `from` into the tx dict it builds, so gas
-        # estimation runs with msg.sender = 0x0 and ERC20 approve reverts with
-        # "approve from the zero address". Disable gas estimation locally so
-        # the hardcoded gas fallback above is used. Remove once the upstream
-        # fix lands.
-        original_gas_estimation = getattr(
-            self.ledger_api, "_is_gas_estimation_enabled", False
+        return executor.execute_transaction(
+            contract=token_contract,
+            method_name=method_name,
+            method_args=method_args,
+            tx_args=tx_args,
         )
-        try:
-            self.ledger_api._is_gas_estimation_enabled = (  # noqa: SLF001  # pylint: disable=protected-access
-                False
-            )
-            return executor.execute_transaction(
-                contract=token_contract,
-                method_name=method_name,
-                method_args=method_args,
-                tx_args=tx_args,
-            )
-        finally:
-            self.ledger_api._is_gas_estimation_enabled = (  # noqa: SLF001  # pylint: disable=protected-access
-                original_gas_estimation
-            )
 
     def get_balance_tracker_address(self) -> str:
         """
--- a/mech_client/domain/payment/token.py
+++ b/mech_client/domain/payment/token.py
@@ -38,6 +38,14 @@ if TYPE_CHECKING:
     from mech_client.domain.execution.base import TransactionExecutor
 
 
+# Client-mode fallback gas for ERC20 approve when estimation fails. Ignored
+# in agent mode (AgentExecutor hardcodes `gas: 0` and Safe estimates its own
+# gas). Sized as a generous buffer above the ~67k measured cost of a
+# 0->nonzero approve on Polygon's Circle USDC (FiatToken) proxy; the historic
+# 60k underfunds that proxy. Re-measure if Circle upgrades the implementation.
+_APPROVE_FALLBACK_GAS = 120_000
+
+
 class TokenPaymentStrategy(PaymentStrategy):
     """Payment strategy for ERC20 token payments (OLAS, USDC).
 
@@ -90,12 +98,13 @@ class TokenPaymentStrategy(PaymentStrategy):
         spender_address: str,
         amount: int,
         executor: Optional["TransactionExecutor"] = None,
-    ) -> Optional[str]:
+    ) -> str:
         """
         Approve token spending for the spender address.
 
         Builds and sends an approve transaction for the specified amount.
-        This must be called before the actual token transfer.
+        This must be called before the actual token transfer. Always sends
+        a transaction and returns the hash; raises if it cannot.
 
         :param payer_address: Address of the payer
         :param spender_address: Address allowed to spend tokens (balance tracker)
@@ -114,10 +123,11 @@ class TokenPaymentStrategy(PaymentStrategy):
         abi = get_abi("IToken.json")
         token_contract = get_contract(token_address, abi, self.ledger_api)
 
-        # Fallback gas if estimation fails. Some tokens (e.g. Circle USDC
-        # proxy on Polygon) need ~67k for a 0->nonzero approve; 60k
-        # underfunds it.
-        tx_args = {"sender_address": payer_address, "value": 0, "gas": 120000}
+        tx_args = {
+            "sender_address": payer_address,
+            "value": 0,
+            "gas": _APPROVE_FALLBACK_GAS,
+        }
         method_name = "approve"
         method_args = {"_to": spender_address, "_value": amount}
 
--- a/mech_client/services/marketplace_service.py
+++ b/mech_client/services/marketplace_service.py
@@ -245,7 +245,15 @@ class MarketplaceService(
         # Handle payment (approval if needed)
         if not use_prepaid and payment_type.is_token():
             balance_tracker = payment_strategy.get_balance_tracker_address()
-            price = self.mech_config.price * len(prompts)
+            # The marketplace pulls exactly `deliveryRate * numRequests` from
+            # the requester via `BalanceTracker.checkAndRecordDeliveryRates`.
+            # Use the per-mech `max_delivery_rate` (read from the mech
+            # contract) instead of the static chain-wide `mech_config.price`
+            # from mechs.json — the latter is a default that does not match
+            # per-mech pricing and under-funds the approve whenever a mech's
+            # rate exceeds it, making the marketplace's transferFrom revert
+            # with "ERC20: transfer amount exceeds allowance".
+            price = max_delivery_rate * len(prompts)
 
             # Check balance
             logger.info(f"Checking {payment_type.name} token balance...")
@@ -279,17 +287,36 @@ class MarketplaceService(
                     )
             logger.info("Token approval complete")
 
-        # Send on-chain marketplace request
+        # Send on-chain marketplace request.
+        # Workaround for valory-xyz/open-aea#924: disable gas estimation for
+        # this transaction too. EthereumApi.build_transaction does not
+        # propagate `from`, so the gas-estimation simulation runs as
+        # msg.sender=0x0 and the balance tracker's transferFrom reverts with
+        # "ERC20: transfer amount exceeds allowance" (allowance[0x0] is 0),
+        # even though the real signed tx would succeed. The fallback gas
+        # limit (mech_config.gas_limit, ~3M on polygon) covers a typical
+        # marketplace request. Remove once the upstream fix lands.
         logger.info("Submitting marketplace request transaction...")
-        tx_hash = self._send_marketplace_request(
-            marketplace_contract=marketplace_contract,
-            data_hashes=data_hashes,
-            max_delivery_rate=max_delivery_rate,
-            payment_type=payment_type,
-            priority_mech=priority_mech_address,
-            response_timeout=response_timeout,
-            use_prepaid=use_prepaid,
+        original_gas_estimation = getattr(
+            self.ledger_api, "_is_gas_estimation_enabled", False
         )
+        try:
+            self.ledger_api._is_gas_estimation_enabled = (  # noqa: SLF001  # pylint: disable=protected-access
+                False
+            )
+            tx_hash = self._send_marketplace_request(
+                marketplace_contract=marketplace_contract,
+                data_hashes=data_hashes,
+                max_delivery_rate=max_delivery_rate,
+                payment_type=payment_type,
+                priority_mech=priority_mech_address,
+                response_timeout=response_timeout,
+                use_prepaid=use_prepaid,
+            )
+        finally:
+            self.ledger_api._is_gas_estimation_enabled = (  # noqa: SLF001  # pylint: disable=protected-access
+                original_gas_estimation
+            )
         tx_url = self.mech_config.transaction_url.format(transaction_digest=tx_hash)
         logger.info(f"Transaction submitted: {tx_url}")
 
--- a/mech_client/services/marketplace_service.py
+++ b/mech_client/services/marketplace_service.py
@@ -243,7 +243,13 @@ class MarketplaceService(
         logger.info(f"Uploaded {len(data_hashes)} metadata hash(es) to IPFS")
 
         # Handle payment (approval if needed)
-        if not use_prepaid and payment_type.is_token():
+        # `is_token()` also matches TOKEN_NVM_USDC, which the factory routes
+        # to NVMPaymentStrategy (no ERC20 approve) — gate on the exact types
+        # that resolve to TokenPaymentStrategy, mirroring factory.py:59.
+        if not use_prepaid and payment_type in (
+            PaymentType.OLAS_TOKEN,
+            PaymentType.USDC_TOKEN,
+        ):
             balance_tracker = payment_strategy.get_balance_tracker_address()
             # The marketplace pulls up to `max_delivery_rate * numRequests`
             # from the requester via `BalanceTracker.checkAndRecordDeliveryRates`.
--- a/mech_client/services/marketplace_service.py
+++ b/mech_client/services/marketplace_service.py
@@ -266,7 +266,7 @@ class MarketplaceService(
 
             # Approve if needed
             logger.info("Sending token approval transaction...")
-            approval_tx_hash = payment_strategy.approve_if_needed(
+            approval_tx_hash = payment_strategy.approve_if_needed(  # pylint: disable=assignment-from-none
                 payer_address=sender,
                 spender_address=balance_tracker,
                 amount=price,
@@ -287,36 +287,16 @@ class MarketplaceService(
                     )
             logger.info("Token approval complete")
 
-        # Send on-chain marketplace request.
-        # Workaround for valory-xyz/open-aea#924: disable gas estimation for
-        # this transaction too. EthereumApi.build_transaction does not
-        # propagate `from`, so the gas-estimation simulation runs as
-        # msg.sender=0x0 and the balance tracker's transferFrom reverts with
-        # "ERC20: transfer amount exceeds allowance" (allowance[0x0] is 0),
-        # even though the real signed tx would succeed. The fallback gas
-        # limit (mech_config.gas_limit, ~3M on polygon) covers a typical
-        # marketplace request. Remove once the upstream fix lands.
         logger.info("Submitting marketplace request transaction...")
-        original_gas_estimation = getattr(
-            self.ledger_api, "_is_gas_estimation_enabled", False
+        tx_hash = self._send_marketplace_request(
+            marketplace_contract=marketplace_contract,
+            data_hashes=data_hashes,
+            max_delivery_rate=max_delivery_rate,
+            payment_type=payment_type,
+            priority_mech=priority_mech_address,
+            response_timeout=response_timeout,
+            use_prepaid=use_prepaid,
         )
-        try:
-            self.ledger_api._is_gas_estimation_enabled = (  # noqa: SLF001  # pylint: disable=protected-access
-                False
-            )
-            tx_hash = self._send_marketplace_request(
-                marketplace_contract=marketplace_contract,
-                data_hashes=data_hashes,
-                max_delivery_rate=max_delivery_rate,
-                payment_type=payment_type,
-                priority_mech=priority_mech_address,
-                response_timeout=response_timeout,
-                use_prepaid=use_prepaid,
-            )
-        finally:
-            self.ledger_api._is_gas_estimation_enabled = (  # noqa: SLF001  # pylint: disable=protected-access
-                original_gas_estimation
-            )
         tx_url = self.mech_config.transaction_url.format(transaction_digest=tx_hash)
         logger.info(f"Transaction submitted: {tx_url}")
 
--- a/mech_client/services/marketplace_service.py
+++ b/mech_client/services/marketplace_service.py
@@ -21,7 +21,7 @@
 
 import logging
 from dataclasses import dataclass
-from typing import Any, Dict, FrozenSet, List, Optional, Tuple
+from typing import Any, Dict, FrozenSet, List, Optional, Tuple, cast
 
 import requests
 from aea_ledger_ethereum import EthereumCrypto
@@ -245,8 +245,8 @@ class MarketplaceService(
         # Handle payment (approval if needed)
         if not use_prepaid and payment_type.is_token():
             balance_tracker = payment_strategy.get_balance_tracker_address()
-            # The marketplace pulls exactly `deliveryRate * numRequests` from
-            # the requester via `BalanceTracker.checkAndRecordDeliveryRates`.
+            # The marketplace pulls up to `max_delivery_rate * numRequests`
+            # from the requester via `BalanceTracker.checkAndRecordDeliveryRates`.
             # Use the per-mech `max_delivery_rate` (read from the mech
             # contract) instead of the static chain-wide `mech_config.price`
             # from mechs.json — the latter is a default that does not match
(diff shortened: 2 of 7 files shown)


mech_client/domain/payment/token.py
```diff
@@ -84,7 +84,7 @@ def check_balance(
         balance = token_contract.functions.balanceOf(payer_address).call()
         return balance >= amount
 
-    def approve_if_needed(  # pylint: disable=too-many-locals
+    def approve_if_needed(
```
OjusWiZard, line 95: Consider narrowing the override return type to `-> str`. This method never returns `None` (it raises, or returns `executor.execute_transaction() -> str`); only `NativePaymentStrategy` returns `None`, and that path is unreachable here behind `is_token()`. Narrowing lets you drop the `# pylint: disable=assignment-from-none` and the dead `if approval_tx_hash:` guard at the call site.

mech_client/domain/payment/token.py
```diff
@@ -114,8 +114,10 @@ def approve_if_needed(  # pylint: disable=too-many-locals
         abi = get_abi("IToken.json")
         token_contract = get_contract(token_address, abi, self.ledger_api)
 
-        # Build approval transaction arguments
-        tx_args = {"sender_address": payer_address, "value": 0, "gas": 60000}
+        # Fallback gas if estimation fails. Some tokens (e.g. Circle USDC
+        # proxy on Polygon) need ~67k for a 0->nonzero approve; 60k
+        # underfunds it.
+        tx_args = {"sender_address": payer_address, "value": 0, "gas": 120000}
```
OjusWiZard: Comment/value mismatch: it cites "~67k ... 60k underfunds it" but sets `120000`, with no rationale for the ~2x jump. Rot-prone — a future reader trusting the ~67k figure might trim it. Suggest adding e.g. "120k is a generous buffer above the ~67k measured cost."

mech_client/services/marketplace_service.py
```diff
@@ -245,7 +245,15 @@ async def send_request(  # pylint: disable=too-many-arguments,too-many-locals
         # Handle payment (approval if needed)
         if not use_prepaid and payment_type.is_token():
             balance_tracker = payment_strategy.get_balance_tracker_address()
-            price = self.mech_config.price * len(prompts)
+            # The marketplace pulls exactly `deliveryRate * numRequests` from
```
OjusWiZard: Minor wording: "pulls **exactly** `deliveryRate * numRequests`" — the code approves `max_delivery_rate * ...`. "Exactly" only holds if the settled rate equals the signed max. The approve is an upper bound, which is fine; the word "exactly" overstates it.

mech_client/services/marketplace_service.py
```diff
@@ -245,7 +245,15 @@ async def send_request(  # pylint: disable=too-many-arguments,too-many-locals
...
+            # the requester via `BalanceTracker.checkAndRecordDeliveryRates`.
+            # Use the per-mech `max_delivery_rate` (read from the mech
+            # contract) instead of the static chain-wide `mech_config.price`
+            # from mechs.json — the latter is a default that does not match
+            # per-mech pricing and under-funds the approve whenever a mech's
+            # rate exceeds it, making the marketplace's transferFrom revert
+            # with "ERC20: transfer amount exceeds allowance".
+            price = max_delivery_rate * len(prompts)
```
OjusWiZard, line 262: This is the headline cleanup, but it has **no regression guard**. In `test_marketplace_service.py`, both `test_onchain_token_payment_flow` and the new revert test set `mech_config.price = 10**18` and `max_delivery_rate = 10**17` (10x apart) yet only assert `approve_if_needed.assert_called_once()` — never the `amount`. Reverting this line to `mech_config.price * len(prompts)` passes the entire suite. Please assert `approve_if_needed.call_args.kwargs["amount"] == max_delivery_rate * len(prompts)`.

mech_client/services/marketplace_service.py
```diff
@@ -258,15 +266,27 @@ async def send_request(  # pylint: disable=too-many-arguments,too-many-locals
 
             # Approve if needed
             logger.info("Sending token approval transaction...")
-            payment_strategy.approve_if_needed(
+            approval_tx_hash = payment_strategy.approve_if_needed(  # pylint: disable=assignment-from-none
```
OjusWiZard: If `approve_if_needed` is narrowed to `-> str` (see token.py:87), this `# pylint: disable=assignment-from-none` and the `if approval_tx_hash:` guard below both become unnecessary.

mech_client/services/marketplace_service.py
```diff
@@ -258,15 +266,27 @@ async def send_request(  # pylint: disable=too-many-arguments,too-many-locals
...
                 executor=self.executor,
             )
+            # Wait for the approval to be mined before submitting the request.
+            # Otherwise the request transaction is built on the pre-approval
+            # ("latest") nonce while the approval is still pending, so both grab
+            # the same nonce -> "replacement transaction underpriced", and the
+            # allowance may not yet be on-chain when the request pulls payment.
+            if approval_tx_hash:
+                approval_receipt = wait_for_receipt(approval_tx_hash, self.ledger_api)
+                if approval_receipt.get("status") == 0:
```
OjusWiZard: `get("status") == 0` treats a missing/`None` status as success. The sibling approval path uses the safer positive form `receipt["status"] != 1` (`domain/subscription/manager.py:175`). Harmless on Polygon (EIP-658 always sets status) but inconsistent — prefer `!= 1` here (and at line 305) to match.

tests/unit/domain/test_payment_strategies.py
```diff
@@ -169,7 +169,10 @@ def test_approve_if_needed_with_executor(
         assert call_args["method_args"] == {"_to": spender_address, "_value": amount}
         assert call_args["tx_args"]["sender_address"] == payer_address
         assert call_args["tx_args"]["value"] == 0
-        assert call_args["tx_args"]["gas"] == 60000
+        # Approve gas must cover a 0->nonzero approve on Circle's USDC proxy
+        # on Polygon, which measures ~67k. 60k underfunds it and the approve
+        # reverts out of gas.
+        assert call_args["tx_args"]["gas"] >= 67_200
```
OjusWiZard: `>= 67_200` is a fine intent-floor, but it would silently pass a regression to e.g. 80k — well below the deliberately chosen 120k headroom. Consider asserting `== 120_000` (keeping the comment to explain the buffer) to lock the constant.

tests/unit/services/test_marketplace_service.py
```diff
@@ -1055,6 +1055,82 @@ async def test_onchain_reverted_tx_raises_error(
...
+
+        service = _build_service(
+            mock_mech_config, mock_ledger_api_cls, mock_executor_factory
+        )
+
+        mock_strategy = MagicMock()
+        mock_strategy.get_balance_tracker_address.return_value = "0x" + "c" * 40
+        mock_strategy.check_balance.return_value = True
+        # Truthy hash triggers the wait_for_receipt + status check
+        mock_strategy.approve_if_needed.return_value = "0xapprovehash"
```
OjusWiZard, line 1119: Good sad-path coverage. Two gaps worth closing: (1) **happy path isn't pinned** — `wait_for_receipt` is also called for the request tx (line ~304), so deleting the new approval-wait would still pass; add a test asserting `wait_for_receipt` is called with the approval hash *before* `_send_marketplace_request`. (2) the `approve_if_needed -> None` falsy branch (`if approval_tx_hash:`) is untested. Nit: the 5-level nested `with patch.object(...)` could use `contextlib.ExitStack` or decorators.

OjusWiZard (commented):
Automated multi-perspective review. The three fixes are sound and correctly reasoned — no blocking runtime bugs. (A reported `wait_for_receipt` returning `None` -> `AttributeError` was investigated and rejected: `web3.Eth.get_transaction_receipt()` raises `TransactionNotFound` for a pending tx, so the poll loop retries and the 300s `TimeoutError` holds.) The substantive gaps are in test coverage — see inline. Also out of this diff's scope but same root cause: `domain/subscription/manager.py:169` still hardcodes `gas: 60000` for a USDC approve and will fail identically on Circle's Polygon proxy.

mech_client/domain/payment/token.py
DIvyaNautiyal07: Done. Narrowed `TokenPaymentStrategy.approve_if_needed` to `-> str`. Used `cast(str, ...)` at the call site since the base `PaymentStrategy` signature stays `Optional[str]` for Native, which lets the disable and the `if approval_tx_hash:` guard go away.

mech_client/domain/payment/token.py
DIvyaNautiyal07: Done. Comment now says 120k is a generous buffer above the ~67k measured cost.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Done, changed to "up to".

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Good catch, this was the most important gap. Added `assert approve_kwargs["amount"] == max_delivery_rate * len(prompts)` to both `test_onchain_token_payment_flow` and `test_onchain_token_reverted_approve_raises_before_request`. The 10x mismatch between `mech_config.price` (10**18) and `max_delivery_rate` (10**17) means reverting that line now fails both tests.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Done via `cast(str, ...)`. Both the disable and the `if approval_tx_hash:` guard are gone.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Done. Both line 282 (approval) and line 305 (request) use `!= 1` now, matching the subscription sibling.

tests/unit/domain/test_payment_strategies.py
DIvyaNautiyal07: Done. Now asserts `== 120_000` with a comment explaining the buffer.

tests/unit/services/test_marketplace_service.py
DIvyaNautiyal07: (1) Done. Happy path now tracks a `call_order` list and asserts `wait_for_receipt` is invoked with the approval hash before `_send_marketplace_request`. Removing the approval-wait fails the test.

(2) Moot now: the `if approval_tx_hash:` guard is gone (see the token.py:87 thread), so there's no falsy branch to test.

On the nesting nit: left as-is to match the rest of the file's pattern; happy to refactor in a follow-up.

DIvyaNautiyal07:
On the out-of-scope note about `domain/subscription/manager.py:169` hardcoding `gas: 60000` for the USDC approve: agreed, same root cause as the marketplace approve. Will open a separate PR for that since the NVM-subscription flow has its own test surface.

mech_client/services/marketplace_service.py
bennyjo, line 262: Nothing stops this from regressing back to the old amount. The only test on the approve amount (`test_marketplace_service.py:920`) checks the call happened, not the amount — it passes either way. Fixtures already differ (`10**18` vs `10**17`), so one line pins it:

```python
assert mock_strategy.approve_if_needed.call_args.kwargs["amount"] == 10**17
```

mech_client/services/marketplace_service.py
```diff
@@ -245,7 +245,15 @@ async def send_request(  # pylint: disable=too-many-arguments,too-many-locals
...
+            # Use the per-mech `max_delivery_rate` (read from the mech
+            # contract) instead of the static chain-wide `mech_config.price`
+            # from mechs.json — the latter is a default that does not match
+            # per-mech pricing and under-funds the approve whenever a mech's
+            # rate exceeds it, making the marketplace's transferFrom revert
+            # with "ERC20: transfer amount exceeds allowance".
+            price = max_delivery_rate * len(prompts)
 
             # Check balance
             logger.info(f"Checking {payment_type.name} token balance...")
```
bennyjo, line 265: Optional: the user pays the mech's rate but never sees the amount before approving. A `logger.info` of `price` here would surface the spend. Fine to skip.

bennyjo (commented):
Two notes on the client-mode USDC payment fix — one test gap, one optional logging suggestion.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Yep, same fix is in the incoming push. The amount assertion is now in both `test_onchain_token_payment_flow` (the happy path you were looking at) and the new `test_onchain_token_reverted_approve_raises_before_request`. The 10x fixture gap (`mech_config.price = 10**18` vs `max_delivery_rate = 10**17`) means reverting the line fails both.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Added. The approval log line now shows `Approving {price} wei ({max_delivery_rate} per request x {len(prompts)}) for spender {balance_tracker}...` so the spend is visible before the approve goes out.

bennyjo (dismissed):

mech_client/services/marketplace_service.py
OjusWiZard: **Fail-open on a missing/None `status`.** `approval_receipt.get("status") == 0` is `False` when `status` is absent or `None` (`None == 0` is `False`), so a malformed / non-EIP-658 receipt is silently treated as success and the request proceeds against an unapproved allowance — surfacing later as a cryptic `ERC20: transfer amount exceeds allowance`. Low probability on the supported chains (all EIP-658), and the **same pattern already exists at line 305** for the request receipt, so this isn't a regression. Consider failing closed in both places:

```python
if approval_receipt.get("status") != 1:   # treat absent/unexpected as failure
```

mech_client/services/marketplace_service.py
```diff
@@ -258,15 +266,27 @@ async def send_request(  # pylint: disable=too-many-arguments,too-many-locals
...
+            # allowance may not yet be on-chain when the request pulls payment.
+            if approval_tx_hash:
+                approval_receipt = wait_for_receipt(approval_tx_hash, self.ledger_api)
+                if approval_receipt.get("status") == 0:
+                    raise ValueError(
+                        f"Token approval transaction reverted. Hash: {approval_tx_hash}. "
+                        f"The request was not sent. This often means the approval ran "
+                        f"out of gas; check the transaction on the block explorer."
+                    )
             logger.info("Token approval complete")
```
OjusWiZard, line 301: `logger.info("Token approval complete")` sits **outside** the `if approval_tx_hash:` guard. If the hash is ever falsy, this logs success with nothing sent or verified. Currently unreachable (the executors always return a hash or raise), but moving it inside the guard — or adding an `else: raise` — keeps the log honest and makes any future gap noisy instead of silent.

mech_client/services/marketplace_service.py
OjusWiZard: **Advisory — the `Optional[str]` invites a dead guard.** `TokenPaymentStrategy.approve_if_needed` only ever returns `executor.execute_transaction(...) -> str` or raises; it never returns `None`. Narrowing the *override* in `token.py` to `-> str` (the base-class `Optional[str]` stays — `NativePaymentStrategy` legitimately returns `None`) lets you drop both the `if approval_tx_hash:` check and this `# pylint: disable=assignment-from-none`, and makes the unconditional wait — the whole point of this PR — type-enforced.

mech_client/domain/payment/token.py
OjusWiZard: Three small advisory points on this fallback:

1. **Comment scope.** "Fallback gas if estimation fails" overclaims — in agent mode `AgentExecutor` builds its own `gas: 0` and discards this value entirely. Suggest: "Client-mode fallback gas … ignored in agent mode (Safe estimates its own gas)."
2. **Magic number.** The repo defines significant literals as module constants (`HTTP_PAYMENT_REQUIRED`, `_MAX_AUTO_DEPOSIT_RATIO`, `DEFAULT_GAS_LIMIT`). Consider `_APPROVE_FALLBACK_GAS = 120_000` at module level.
3. **Stale risk.** The empirical `~67k` is tied to a specific Circle FiatToken proxy version with no version/date marker — e.g. add "Measured against Circle FiatTokenV2.2 on Polygon; re-measure on USDC proxy upgrade."

tests/unit/domain/test_payment_strategies.py
OjusWiZard: **Lower-bound only.** `>= 67_200` passes for anything down to 67,201, silently eroding the safety margin below the actual `120000` constant. Pin it instead:

```python
assert call_args["tx_args"]["gas"] == 120000
```

or bound both sides (`67_200 <= gas <= 200_000`).

tests/unit/services/test_marketplace_service.py
OjusWiZard, line 1119: This new revert-path test is good. Two coverage gaps remain in the **happy path** (`test_onchain_token_payment_flow`, ~L851-920), worth closing here or there:

1. **The core bugfix is unasserted.** That test sets `mech_config.price = 10**18` and `max_delivery_rate = 10**17` (a deliberate 10× gap) but only checks `approve_if_needed.assert_called_once()` — never the `amount`. Reverting `price = max_delivery_rate * len(prompts)` back to `mech_config.price * len(prompts)` leaves every test green. Add:
```python
call_kwargs = mock_strategy.approve_if_needed.call_args[1]
assert call_kwargs["amount"] == 10**17 * len(prompts)
```
2. **The new wait isn't pinned on success.** Nothing asserts `wait_for_receipt` was called with the approval hash, so deleting that call wouldn't fail a test: `mock_wait_receipt.assert_any_call("0xapprovehash", service.ledger_api)`. (Optionally also cover the `approve_if_needed` returns-`None` skip branch.)

OjusWiZard (commented):
Automated multi-perspective review (correctness, comments, tests, silent-failures, type-design, simplification). The three core changes are correct and well-motivated — the comments below are one real test gap, one latent robustness issue (shared with the pre-existing receipt check), and advisory polish. None are blocking.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Agreed and done. Both line 282 (approval) and line 305 (request) now use `!= 1`, so a missing/None status fails closed.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Resolved by dropping the `if approval_tx_hash:` guard entirely. With the cast(str, ...) at the call site and the narrowed `TokenPaymentStrategy.approve_if_needed -> str` override, the wait + status check + success log all run unconditionally — no honest-but-unreachable branch to worry about.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Done. Narrowed `TokenPaymentStrategy.approve_if_needed` to `-> str`; the base `PaymentStrategy.approve_if_needed` stays `Optional[str]` for Native. Used `cast(str, ...)` at the call site since mypy/pylint still see the base signature there, which lets both the `if approval_tx_hash:` check and the `# pylint: disable=assignment-from-none` go away, and makes the unconditional wait type-enforced.

mech_client/domain/payment/token.py
DIvyaNautiyal07: All three done. Extracted `_APPROVE_FALLBACK_GAS = 120_000` at module level, and the comment now says: client-mode fallback only (agent mode's `AgentExecutor` hardcodes `gas: 0` and Safe estimates its own), ~67k measured against Polygon's Circle USDC (FiatToken) proxy, re-measure if Circle upgrades the implementation. Test asserts against the constant now too.

tests/unit/domain/test_payment_strategies.py
DIvyaNautiyal07: Done. Test now asserts `== _APPROVE_FALLBACK_GAS` (the module constant), so an 80k regression that still sits above the measured ~67k cost but eats the buffer fails this assertion.

tests/unit/services/test_marketplace_service.py
DIvyaNautiyal07: Both happy-path gaps closed in the latest push.

(1) `test_onchain_token_payment_flow` now asserts `approve_kwargs["amount"] == max_delivery_rate * 1` (the same 10x fixture gap surfaces a revert of `price = max_delivery_rate * len(prompts)` back to `mech_config.price * len(prompts)`).

(2) Same test tracks a `call_order` list and asserts the approval-hash `wait_for_receipt` runs before `_send_marketplace_request`, so deleting that wait fails the test.

On the optional `approve_if_needed -> None` skip branch: that branch is gone now — `TokenPaymentStrategy.approve_if_needed` is narrowed to `-> str` and the `if approval_tx_hash:` guard at the call site is dropped (see the token.py:87 / marketplace_service.py:269 threads), so nothing to test on that path.

OjusWiZard (commented):
Follow-up review of 571c3e2 (the push addressing the earlier comments). The fixes are correct and the restructured tests pass (90/90), but the guard removal introduced one reachable regression, plus two minor test-quality notes.

mech_client/services/marketplace_service.py
```diff
@@ -245,7 +245,15 @@ async def send_request(  # pylint: disable=too-many-arguments,too-many-locals
         # Handle payment (approval if needed)
         if not use_prepaid and payment_type.is_token():
```
OjusWiZard: **Regression introduced by dropping the `if approval_tx_hash:` guard.** This gate also admits `TOKEN_NVM_USDC` — `is_token()` returns `True` for it (payment_config.py:67-72) — which the factory routes to `NVMPaymentStrategy` (factory.py:62-63), whose `approve_if_needed` returns **`None`** (subscriptions need no approval). With the guard gone, `cast(str, None)` is a runtime no-op, so `approval_tx_hash = None` now flows unguarded into `wait_for_receipt(None, …)` at L289 → polls a `None` hash for the full 300s timeout, then raises a misleading `TimeoutError`.

Reachable today via `mechx request <prompt> <tool> --priority-mech <nvm-usdc-mech>` without `--prepaid`/`--use-offchain` (`request_cmd.py:184` leaves `use_prepaid=False`). The old guard degraded gracefully here (skip approval → send request); this hangs for 5 minutes. The `cast(str, …)` is dishonest for this type.

**Fix** — narrow the gate to the types that actually use `TokenPaymentStrategy` (mirrors the factory and the existing `_SUPPORTED_DEPOSIT_PAYMENT_TYPES` set):
```python
if not use_prepaid and payment_type in (PaymentType.OLAS_TOKEN, PaymentType.USDC_TOKEN):
```
That restores correct NVM handling and makes the `cast` honest (only pure-token strategies reach it).

tests/unit/services/test_marketplace_service.py
```diff
@@ -895,29 +898,46 @@ async def test_onchain_token_payment_flow(
...
                             prompts=("hello",),
                             tools=("some-tool",),
                             use_prepaid=False,
                         )
 
-        # Verify approval was triggered
+        # Verify approval was triggered with the per-mech rate, not mech_config.price
         mock_strategy.approve_if_needed.assert_called_once()
+        approve_kwargs = mock_strategy.approve_if_needed.call_args.kwargs
+        assert approve_kwargs["amount"] == max_delivery_rate * 1
```
OjusWiZard, line 933: **The `* len(prompts)` multiplier isn't actually guarded.** Both flow tests use a single prompt, so `max_delivery_rate * 1` is numerically just `max_delivery_rate` — dropping `* len(prompts)` at `marketplace_service.py:256` would still pass here (and at L1154). No token-approval test uses >1 prompt. Add a 2-prompt case asserting `== max_delivery_rate * 2`, or parametrize over `n_prompts in (1, 2)`.

tests/unit/services/test_marketplace_service.py
```diff
@@ -886,6 +886,9 @@ async def test_onchain_token_payment_flow(
 
         mock_contract = MagicMock()
         mock_push_metadata.return_value = ("0x" + "b" * 64, "ipfs://hash")
+        # Approve returns a distinct hash so we can verify wait_for_receipt
+        # is called with it before the request transaction is built.
+        mock_strategy.approve_if_needed.return_value = "0xapprovehash"
         mock_wait_receipt.return_value = {"status": 1}
```
OjusWiZard: Dead line: `mock_wait_receipt.side_effect` is assigned a few lines below, and in `unittest.mock` a set `side_effect` fully overrides `return_value`, so this is never consulted. Safe to remove — the lambda already returns `{"status": 1}` for both wait calls.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Caught and fixed. You're exactly right on the routing — `is_token()` includes `TOKEN_NVM_USDC` (USDC under the hood) but the factory sends that to `NVMPaymentStrategy` whose `approve_if_needed` returns `None`, and after dropping the guard the cast was a runtime no-op so `None` was reaching `wait_for_receipt` and burning the full 300s.

Applied your fix: the gate is now `payment_type in (PaymentType.OLAS_TOKEN, PaymentType.USDC_TOKEN)`, mirroring `factory.py:59`. That restores the pre-PR degrade-gracefully behaviour for NVM (skip approve, send request) and makes the `cast(str, ...)` honest since only the two ERC20-approval strategies reach it. Added `test_onchain_token_nvm_usdc_skips_approve_branch` that exercises the NVM USDC path and asserts (a) `approve_if_needed` is not called and (b) no None/empty hash is passed to `wait_for_receipt`.

tests/unit/services/test_marketplace_service.py
DIvyaNautiyal07: Good catch. Added `test_onchain_token_approve_amount_scales_with_prompt_count` which uses 2 prompts and asserts `approve_kwargs["amount"] == max_delivery_rate * len(prompts) == max_delivery_rate * 2`. Dropping `* len(prompts)` at line 256 now fails the new test (the existing single-prompt tests would still pass either way, as you noted).

tests/unit/services/test_marketplace_service.py
DIvyaNautiyal07: Removed. The lambda set on `side_effect` covers both wait calls (approval + request) with `{"status": 1}`.

OjusWiZard (approved):
Approving. Reviewed across the full series — base PR (657bff2), the address-comments push (571c3e2), and the NVM gate fix (efd0f1f).

The three client-mode USDC fixes are correct: per-mech `max_delivery_rate` approve amount, wait-for-approval-receipt before the request, and the 120k fallback gas constant. All earlier review threads were addressed, and the regression that the refactor briefly introduced (`is_token()` admitting TOKEN_NVM_USDC into the approve branch → a None hash polled to timeout) is closed by gating on the exact ERC20 types that route to TokenPaymentStrategy, with a guarding regression test plus a multi-prompt amount-scaling test.

Two non-blocking advisory notes remain (the gate tuple duplicates factory.py:59 — a shared frozenset would remove drift risk; and the NVM test could assert the request was still sent), but neither needs to block merge.

bennyjo (approved):
LGTM. The three fixes are sound and the prior review round is fully addressed — fail-closed receipt checks (`!= 1`), the `-> str` narrowing with the dead guard removed, the `_APPROVE_FALLBACK_GAS` constant, and discriminating tests for the approve amount, the approval-before-request ordering, and the NVM-skip path. I verified the exact-type gate against `factory.py:59`: only `OLAS_TOKEN`/`USDC_TOKEN` reach `TokenPaymentStrategy`, so the `cast(str, ...)` is honest and the NVM 300s-timeout regression is genuinely closed.

Two non-blocking follow-ups for later:
- `domain/subscription/manager.py:169` still hardcodes `gas: 60000` for the Base USDC approve — same root cause as this PR; ideally reuse `_APPROVE_FALLBACK_GAS` so the two approve paths can't drift. It fails loudly (receipt status is checked), so lower urgency.
- Approve amount now comes from the mech's on-chain `maxDeliveryRate()` rather than the static `mech_config.price`, removing the de-facto ceiling. The new spend log covers visibility; a sanity bound could be a future hardening if non-interactive clients are a concern.

PR #241, opened 2026-06-17
feat(offchain): structured 402 payment handling + auto-deposit [WS-2]

DhairyaPatel7:
## WS-2 — mech-client offchain payment handling

Part of the Marketplace/API offchain migration (Thread 1, requester side). Targets the offchain server contract defined in **mech#454** (structured 402 + `Payment-Receipt`).

### Context
The offchain request path already exists on `main`: it skips the IPFS upload, computes the content CID locally, POSTs the signed request to `/send_signed_requests`, and polls `/fetch_offchain_info`. The missing WS-2 piece is the **payment handshake** — today any non-2xx from the mech is collapsed into a generic `"rejected"` error, with no handling of the structured 402 the mech now returns.

### What this adds
- **Parse the structured 402 challenge** (`required` / `currentBalance` / `payTo` / `asset` / `chainId` / `error`) + the `WWW-Authenticate: Payment` header.
- **Without auto-deposit** (default): raise an actionable error telling the requester exactly how much to deposit and where — instead of an opaque rejection.
- **With `--auto-deposit`**: deposit the shortfall into the prepaid balance via `DepositService` (native / OLAS / USDC; NVM subscription types are explicitly unsupported since they don't use a balance-tracker deposit), then retry the request once.
- **Log the `Payment-Receipt`** header returned on a 200.
- New `--auto-deposit` CLI flag (default off), threaded through `send_request`.
- **Bit-for-bit CID parity with the mech** (resolves Benny's flag on mech#454). The previous offchain CID derivation went through `IPFSClient.upload(file, pin=False)`, which still contacts a real IPFS daemon and depends on the daemon's encoding. Replaced with a byte-identical port of the mech's `compute_cidv1` (single-block UnixFS + DAG-PB, bare-file CIDv1, pure Python, no daemon). Pinned with a shared fixture set that mirrors the mech's `test_local_cid.py` — every `(content, CID)` pair was reproduced by running `ipfs add --cid-version=1 --raw-leaves=false` and appears in both repos. A regression that re-introduces the daemon path fails the new `test_no_ipfs_daemon_contacted` test loudly.

### Tests
Full suite green; black-clean. New tests cover 402 parse (full + malformed body), receipt logging (present/absent), auto-deposit (native / usdc / olas / no-shortfall skip / unsupported-type raise), the POST paths (direct 200, 402-no-deposit-raises, 402-auto-deposit-retries-to-200), the CID parity fixtures, and the no-daemon-contacted assertion.

### Notes
- Depends on mech#454's 402 contract; not on it being merged.

### DIFF
--- a/mech_client/cli/commands/request_cmd.py
+++ b/mech_client/cli/commands/request_cmd.py
@@ -72,7 +72,8 @@ def _format_delivery_output(delivery_data: Any) -> str:
 )
 @click.option(
     "--auto-deposit",
-    type=bool,
+    is_flag=True,
+    default=False,
     help="On an offchain HTTP 402, top up the prepaid balance and retry.",
 )
 @click.option(
@@ -176,7 +177,6 @@ def request(
     # Process flags
     use_offchain = use_offchain or False
     use_prepaid = use_prepaid or use_offchain
-    auto_deposit = auto_deposit or False
 
     # Validate tools
     if not tools:
--- a/mech_client/cli/commands/request_cmd.py
+++ b/mech_client/cli/commands/request_cmd.py
@@ -71,6 +71,11 @@ def _format_delivery_output(delivery_data: Any) -> str:
     help="Use offchain mech (URL discovered from on-chain metadata).",
 )
 @click.option(
+    "--auto-deposit",
+    type=bool,
+    help="On an offchain HTTP 402, top up the prepaid balance and retry.",
+)
+@click.option(
     "--tools",
     type=str,
     multiple=True,
@@ -109,6 +114,7 @@ def request(
     chain_config: str,
     use_prepaid: bool,
     use_offchain: bool,
+    auto_deposit: bool,
     key: Optional[str],
     tools: Optional[tuple],
     extra_attribute: Optional[List[str]] = None,
@@ -138,6 +144,7 @@ def request(
     :param chain_config: Chain configuration name (gnosis, base, polygon, optimism).
     :param use_prepaid: Use the marketplace prepaid balance instead of per-request payment.
     :param use_offchain: Route delivery off-chain (auto-discovered URL); implies prepaid.
+    :param auto_deposit: On an offchain HTTP 402, top up the prepaid balance and retry.
     :param key: Optional path to a private-key file (required in client mode).
     :param tools: One or more tool names matching ``prompts`` positionally.
     :param extra_attribute: Optional ``key=value`` extras attached to each request.
@@ -169,6 +176,7 @@ def request(
     # Process flags
     use_offchain = use_offchain or False
     use_prepaid = use_prepaid or use_offchain
+    auto_deposit = auto_deposit or False
 
     # Validate tools
     if not tools:
@@ -204,6 +212,7 @@ def request(
             priority_mech=priority_mech,
             use_prepaid=use_prepaid,
             use_offchain=use_offchain,
+            auto_deposit=auto_deposit,
             extra_attributes=extra_attributes_dict,
             timeout=timeout,
         )
--- a/mech_client/cli/commands/request_cmd.py
+++ b/mech_client/cli/commands/request_cmd.py
@@ -74,7 +74,12 @@ def _format_delivery_output(delivery_data: Any) -> str:
     "--auto-deposit",
     is_flag=True,
     default=False,
-    help="On an offchain HTTP 402, top up the prepaid balance and retry.",
+    help=(
+        "On an offchain HTTP 402, top up the prepaid balance and retry. "
+        "Trusts the mech's reported shortfall, capped at 10x the signed "
+        "max_delivery_rate; above that the request is refused so the user "
+        "can deposit manually."
+    ),
 )
 @click.option(
     "--tools",
--- a/mech_client/infrastructure/ipfs/metadata.py
+++ b/mech_client/infrastructure/ipfs/metadata.py
@@ -25,10 +25,8 @@ import tempfile
 import uuid
 from typing import Any, Dict, Optional, Tuple
 
-import multibase
-import multicodec
 from mech_client.infrastructure.ipfs.client import IPFSClient
-from mech_client.infrastructure.ipfs.local_cid import compute_cidv1
+from mech_client.infrastructure.ipfs.local_cid import compute_cidv1_bytes
 
 
 def fetch_ipfs_hash(
@@ -65,15 +63,13 @@ def fetch_ipfs_hash(
     # the request is rejected.
     ipfs_data = json.dumps(metadata)
 
-    # Compute the CIDv1 in-process. ``compute_cidv1`` already returns a
-    # base32-lower multibase CIDv1 (``bafy...``); convert to the legacy
-    # ``f01...`` hex form the downstream truncation math expects (multibase
-    # 'f' + version '01' + dag-pb codec '70' + sha256 multihash code '12' +
-    # length '20' + 32 hex bytes of digest).
-    cidv1 = compute_cidv1(ipfs_data.encode("utf-8"))
-    cid_bytes = multibase.decode(cidv1)
-    multihash_bytes = multicodec.remove_prefix(cid_bytes)
-    v1_file_hash_hex = "f01" + multihash_bytes.hex()
+    # Compute the CIDv1 in-process and emit the legacy ``f01...`` hex form
+    # the downstream truncation math expects (multibase 'f' + version '01' +
+    # dag-pb codec '70' + sha256 multihash code '12' + length '20' + 32 hex
+    # bytes of digest). ``compute_cidv1_bytes`` returns the raw CID bytes so
+    # we can hex-encode directly — no multibase / multicodec round-trip.
+    cid_bytes = compute_cidv1_bytes(ipfs_data.encode("utf-8"))
+    v1_file_hash_hex = "f" + cid_bytes.hex()
 
     # Truncate hash for on-chain use (remove first 9 chars and add 0x prefix)
     truncated_hash = "0x" + v1_file_hash_hex[9:]
--- a/mech_client/infrastructure/ipfs/metadata.py
+++ b/mech_client/infrastructure/ipfs/metadata.py
@@ -25,7 +25,10 @@ import tempfile
 import uuid
 from typing import Any, Dict, Optional, Tuple
 
+import multibase
+import multicodec
 from mech_client.infrastructure.ipfs.client import IPFSClient
+from mech_client.infrastructure.ipfs.local_cid import compute_cidv1
 
 
 def fetch_ipfs_hash(
@@ -34,11 +37,15 @@ def fetch_ipfs_hash(
     extra_attributes: Optional[Dict[str, Any]] = None,
 ) -> Tuple[str, str, str]:
     """
-    Create metadata and compute IPFS hash without uploading.
+    Build offchain request metadata and compute its content CID locally.
 
-    Used for offchain requests where the offchain mech handles IPFS upload.
-    Creates a JSON metadata file, computes its IPFS hash, and returns both
-    the hash and the metadata content.
+    Used for offchain requests where the offchain mech never publishes the
+    content to public IPFS. The CID is computed entirely in-process via
+    ``compute_cidv1`` — no IPFS daemon is contacted, so the prompt and any
+    extra attributes never leave the requester's process. The returned CID
+    is byte-identical to what the mech computes locally on receipt, so the
+    requester's signature over the CID verifies against the mech's
+    recomputed value.
 
     :param prompt: Prompt string
     :param tool: Tool string
@@ -52,23 +59,25 @@ def fetch_ipfs_hash(
     if extra_attributes:
         metadata.update(extra_attributes)
 
-    # Convert metadata to JSON string for offchain transmission
+    # Convert metadata to JSON string for offchain transmission. The exact
+    # bytes hashed below MUST match the bytes the mech receives in ipfs_data,
+    # otherwise the mech's local recomputation produces a different CID and
+    # the request is rejected.
     ipfs_data = json.dumps(metadata)
 
-    dirpath = tempfile.mkdtemp()
-    try:
-        file_name = dirpath + "/metadata.json"
-        with open(file_name, "w", encoding="utf-8") as f:
-            json.dump(metadata, f)
+    # Compute the CIDv1 in-process. ``compute_cidv1`` already returns a
+    # base32-lower multibase CIDv1 (``bafy...``); convert to the legacy
+    # ``f01...`` hex form the downstream truncation math expects (multibase
+    # 'f' + version '01' + dag-pb codec '70' + sha256 multihash code '12' +
+    # length '20' + 32 hex bytes of digest).
+    cidv1 = compute_cidv1(ipfs_data.encode("utf-8"))
+    cid_bytes = multibase.decode(cidv1)
+    multihash_bytes = multicodec.remove_prefix(cid_bytes)
+    v1_file_hash_hex = "f01" + multihash_bytes.hex()
 
-        client = IPFSClient()
-        _, v1_file_hash_hex = client.upload(file_name, pin=False)
-
-        # Truncate hash for on-chain use (remove first 9 chars and add 0x prefix)
-        truncated_hash = "0x" + v1_file_hash_hex[9:]
-        return truncated_hash, v1_file_hash_hex, ipfs_data
-    finally:
-        shutil.rmtree(dirpath, ignore_errors=True)
+    # Truncate hash for on-chain use (remove first 9 chars and add 0x prefix)
+    truncated_hash = "0x" + v1_file_hash_hex[9:]
+    return truncated_hash, v1_file_hash_hex, ipfs_data
 
 
 def push_metadata_to_ipfs(
--- a/mech_client/services/marketplace_service.py
+++ b/mech_client/services/marketplace_service.py
@@ -42,6 +42,12 @@ from web3.contract import Contract as Web3Contract
 
 logger = logging.getLogger(__name__)
 
+# HTTP status the offchain mech returns when the requester's prepaid balance is
+# insufficient (structured 402 challenge, see mech task_execution handlers).
+HTTP_PAYMENT_REQUIRED = 402
+# Outbound HTTP timeout (seconds) for the offchain request POST.
+OFFCHAIN_HTTP_TIMEOUT = 30
+
 
 class MarketplaceService(
     BaseTransactionService
@@ -90,6 +96,7 @@ class MarketplaceService(
         priority_mech: Optional[str] = None,
         use_prepaid: bool = False,
         use_offchain: bool = False,
+        auto_deposit: bool = False,
         extra_attributes: Optional[Dict[str, Any]] = None,
         timeout: Optional[float] = None,
     ) -> Dict[str, Any]:
@@ -101,6 +108,8 @@ class MarketplaceService(
         :param priority_mech: Priority mech address (optional)
         :param use_prepaid: Use prepaid balance instead of per-request payment
         :param use_offchain: Use offchain mech (URL discovered from metadata)
+        :param auto_deposit: On an offchain HTTP 402, top up the prepaid balance
+            with the shortfall and retry once (only applies to the offchain path)
         :param extra_attributes: Extra attributes for metadata
         :param timeout: Timeout for delivery watching
         :return: Dictionary with request results
@@ -147,6 +156,7 @@ class MarketplaceService(
                 mech_offchain_url=offchain_url,
                 extra_attributes=extra_attributes,
                 timeout=timeout or 300.0,
+                auto_deposit=auto_deposit,
             )
 
         # On-chain flow
@@ -243,6 +253,7 @@ class MarketplaceService(
         mech_offchain_url: str,
         extra_attributes: Optional[Dict[str, Any]],
         timeout: float,
+        auto_deposit: bool = False,
     ) -> Dict[str, Any]:
         """
         Send offchain request to mech HTTP endpoint.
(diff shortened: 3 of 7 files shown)


mech_client/services/marketplace_service.py
```diff
@@ -355,6 +364,126 @@ async def _send_offchain_request(  # pylint: disable=too-many-arguments,too-many
...
+                    f"top up your prepaid balance. ({challenge['error']})"
+                )
+            self._auto_deposit_for_402(payment_type, challenge)
+            response = requests.post(
+                url=url,
+                data=payload,
+                headers={"Content-Type": "application/json"},
+                timeout=OFFCHAIN_HTTP_TIMEOUT,
+            )
+        self._log_payment_receipt(response)
```
OjusWiZard, line 508: **Important — a second 402 after auto-deposit is returned silently as if successful.**

The retry POST on L405-410 has no status check before `return response`. If the mech still 402s (deposit too small, tx not yet mined, price moved, or shortfall computed as 0 — see the `_parse_402_challenge` comment), this 402 flows back to `_send_offchain_request`, where `if not response.ok` raises the *generic* path. A 402 body has no `reason` key, so the user sees:

```
Offchain request rejected: Payment Required (HTTP 402)
```

…**after their deposit tx already confirmed** (`deposit_*` waits for receipt) — no amounts, no hint a deposit happened, no refund.

```python
response = requests.post(...)
if response.status_code == HTTP_PAYMENT_REQUIRED:
    retry = self._parse_402_challenge(response)
    raise ValueError(
        f"Auto-deposit did not clear the shortfall: still need {retry['required']} "
        f"(balance {retry['current_balance']}). The deposit may not be mined yet "
        f"or the price changed. ({retry['error']})"
    )
```

The `[402, 402]` path is also untested (only `[402, 200]` is covered) — please add a test asserting this error.

Minor, same method: the two identical `requests.post(...)` calls (L388-393 and L405-410) could be a small `_do_post(url, payload)` helper.

mech_client/services/marketplace_service.py
```diff
@@ -355,6 +364,126 @@ async def _send_offchain_request(  # pylint: disable=too-many-arguments,too-many
...
+        """
+        authenticate = response.headers.get("WWW-Authenticate")
+        if authenticate:
+            logger.info(f"Mech payment challenge: {authenticate}")
+        try:
+            body = response.json()
+        except ValueError:
+            body = {}
+        return {
+            "required": int(body.get("required", 0) or 0),
```
OjusWiZard: **Important — malformed 402 bodies fail confusingly.**

Two edges here:

1. **Non-JSON body** → `except ValueError: body = {}` (L427-428) zeroes the whole challenge. With `--auto-deposit`, `shortfall` becomes 0 so `_auto_deposit_for_402` silently no-ops, then the pointless retry fires (feeds the silent-second-402 issue). Without it, the message reads `need 0 (current balance 0) deposited to ''`.
2. **Non-numeric value** (e.g. `"required": "N/A"`) → `int(body.get("required", 0) or 0)` raises an **unhandled** `ValueError` and escapes as a raw traceback. The `or 0` only guards falsy values, not bad strings.

Suggest a small helper and a warning on the fallback:

```python
def _safe_int(val: Any, default: int = 0) -> int:
    try:
        return int(val) if val is not None else default
    except (ValueError, TypeError):
        return default
```
and `logger.warning("402 body not valid JSON: %r", response.text[:500])` before `body = {}`.

mech_client/services/marketplace_service.py
```diff
@@ -355,6 +364,126 @@ async def _send_offchain_request(  # pylint: disable=too-many-arguments,too-many
...
+                url=url,
+                data=payload,
+                headers={"Content-Type": "application/json"},
+                timeout=OFFCHAIN_HTTP_TIMEOUT,
+            )
+        self._log_payment_receipt(response)
+        return response
+
+    @staticmethod
+    def _parse_402_challenge(response: requests.Response) -> Dict[str, Any]:
```
OjusWiZard: **Advisory (type design)** — the challenge is `Dict[str, Any]`, consumed by string-key access in three places (`_auto_deposit_for_402`, the no-deposit error message, the retry). No static guarantee the six keys exist or that `required`/`current_balance` are ints; a key typo silently returns `Any` then `KeyError`s at runtime. A `@dataclass(frozen=True) PaymentChallenge` with typed fields and a `shortfall` property would make field access checkable and dedupe the `required - current_balance` math.

mech_client/services/marketplace_service.py
```diff
@@ -355,6 +364,126 @@ async def _send_offchain_request(  # pylint: disable=too-many-arguments,too-many
...
+            agent_mode=self.agent_mode,
+            crypto=self.crypto,
+            safe_address=self.safe_address,
+            ethereum_client=self.ethereum_client,
+        )
+        logger.info(f"Auto-depositing {shortfall} to top up the prepaid balance...")
+        # Explicit per-type dispatch (not is_native/is_token, which group the NVM
+        # subscription types in — those use a subscription, not a balance-tracker
+        # deposit, so auto-deposit doesn't apply to them).
+        if payment_type == PaymentType.NATIVE:
```
OjusWiZard, line 614: **Advisory** — this `if/elif` over `PaymentType` falls to `else` for any future enum member, and the intent (NVM types are deliberately excluded) lives only in the comment. When a sixth `PaymentType` is added it silently hits the `else` and raises only at runtime on the offchain 402 path. A named `frozenset` of supported types (or a `match` with explicit unsupported cases) makes the supported set visible and easy to keep exhaustive.

mech_client/infrastructure/ipfs/metadata.py
```diff
@@ -52,23 +59,25 @@ def fetch_ipfs_hash(
...
-        return truncated_hash, v1_file_hash_hex, ipfs_data
-    finally:
-        shutil.rmtree(dirpath, ignore_errors=True)
+    # Compute the CIDv1 in-process. ``compute_cidv1`` already returns a
+    # base32-lower multibase CIDv1 (``bafy...``); convert to the legacy
+    # ``f01...`` hex form the downstream truncation math expects (multibase
+    # 'f' + version '01' + dag-pb codec '70' + sha256 multihash code '12' +
+    # length '20' + 32 hex bytes of digest).
+    cidv1 = compute_cidv1(ipfs_data.encode("utf-8"))
+    cid_bytes = multibase.decode(cidv1)
```
OjusWiZard: **Advisory (simplify)** — `compute_cidv1` builds the raw CID bytes, base32-encodes them to a string, and here it's immediately decoded back via `multibase.decode` + `multicodec.remove_prefix`. It's **correct today** (verified: `remove_prefix` strips only the 1-byte version, leaving codec `70`, so the result is `f01701220…`), but the round-trip pulls `multibase`+`multicodec` into the hot path and leans on `remove_prefix`'s exact stripping. Cleaner: expose `compute_cidv1_bytes(content) -> bytes` from `local_cid` and do `v1_file_hash_hex = "f01" + cid_bytes[2:].hex()`, dropping both library calls here.

mech_client/cli/commands/request_cmd.py
```diff
@@ -70,6 +70,11 @@ def _format_delivery_output(delivery_data: Any) -> str:
     type=bool,
     help="Use offchain mech (URL discovered from on-chain metadata).",
 )
+@click.option(
+    "--auto-deposit",
```
OjusWiZard, line 74: **Advisory** — `type=bool` means `--auto-deposit` requires a value: bare `--auto-deposit` errors with `Option '--auto-deposit' requires an argument` (verified); you must write `--auto-deposit true`. This matches the existing `--use-prepaid`/`--use-offchain` convention (so not a regression), but that convention's own usage example at L139 shows `--use-prepaid` as a bare flag, which would fail. Consider `is_flag=True, default=False` (and drop the `auto_deposit or False` guard on L179) for the idiomatic toggle behavior.

tests/unit/services/test_marketplace_service.py
```diff
@@ -1409,3 +1448,179 @@ async def test_send_request_routes_to_offchain(
...
+        """A non-JSON 402 body falls back to safe defaults."""
+        resp = MagicMock()
+        resp.headers = {}
+        resp.json.side_effect = ValueError("no json")
+        challenge = MarketplaceService._parse_402_challenge(resp)
+        assert challenge["required"] == 0
+        assert challenge["current_balance"] == 0
+        assert challenge["error"] == "payment required"
+
+    def test_log_payment_receipt_present(self) -> None:
```
OjusWiZard: **Advisory (test)** — `test_log_payment_receipt_present`/`_absent` assert only that the call doesn't raise; a refactor that drops the receipt logging would still pass. The receipt is a user-visible audit value — assert it's logged:

```python
def test_log_payment_receipt_present_logs_value(self, caplog) -> None:
    resp = _mock_http_response(200, headers={"Payment-Receipt": "abc123"})
    with caplog.at_level(logging.INFO):
        MarketplaceService._log_payment_receipt(resp)
    assert "abc123" in caplog.text
```

tests/unit/services/test_marketplace_service.py
```diff
@@ -1409,3 +1448,179 @@ async def test_send_request_routes_to_offchain(
...
+    def test_post_offchain_request_402_without_auto_deposit_raises(self) -> None:
+        """A 402 without auto-deposit raises an actionable error."""
+        service = _build_offchain_service()
+        with patch(
+            "mech_client.services.marketplace_service.requests.post",
+            return_value=_mock_http_response(
+                402, {"required": "100", "currentBalance": "0", "payTo": "0xbt"}
+            ),
+        ):
+            with pytest.raises(ValueError, match="requires payment"):
```
OjusWiZard: **Advisory (test)** — `match="requires payment"` is broad enough that swapping out the entire actionable message still passes. The error's value is the `required` amount and `pay_to` address it surfaces; tighten the regex to assert those, e.g. `match=r"need 100.*0xbt"`.

OjusWiZard (commented):
Automated multi-perspective review (correctness, tests, silent-failures, type-design, simplification). Inline comments below cover only issues **introduced by this PR**; pre-existing items (the `data=payload` vs `Content-Type: application/json` mismatch, and the broad-except around reason parsing) are intentionally omitted. The happy path and the CID parity tests look solid — comments are concentrated on the 402 error/retry edges.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Fixed in 9742f52. `_post_offchain_request` now checks the retry response and raises an actionable error when the second 402 lands:

```
Auto-deposit did not clear the 402: deposited to 0xbt but the mech still
requires 150 (current balance 100, remaining shortfall 50). The deposit may
not be mined yet, the price may have moved, or the asset may be wrong.
(insufficient balance)
```

The `[402, 402]` path is covered by `test_post_offchain_request_second_402_after_deposit_raises_actionable` — it drives both POSTs through, asserts the deposit was called once, and asserts the message names both `0xbt` and `remaining shortfall 50`. The minor: extracted `_do_offchain_post(url, payload)` so the two POSTs share the headers + timeout.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Fixed in 9742f52. Added a `_safe_int(value, default=0)` helper that catches both `TypeError` and `ValueError` from `int(...)`. All four numeric fields (`required`, `currentBalance`, `chainId`) go through it. The non-JSON branch now logs:

```
WARNING 402 body was not valid JSON; falling back to defaults. Raw body
(first 500 chars): '<html>500</html>'
```

Locked in by `test_parse_402_challenge_non_numeric_value_is_tolerated` (covers `"N/A"` / `None` / `"not-a-number"`) and `test_parse_402_challenge_invalid_body_defaults_and_warns` (asserts the WARNING is logged with the raw body).

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Done in 9742f52. New `@dataclass(frozen=True) PaymentChallenge` with typed fields and a `shortfall = max(0, required - current_balance)` property. Migrated the three call sites (`_auto_deposit_for_402`, the no-deposit error message, the second-402 error message). Key typos at any of them are now mypy errors instead of runtime `KeyError`s, and the shortfall math is in one place.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Done in 9742f52. New `_SUPPORTED_DEPOSIT_PAYMENT_TYPES` frozenset declared at module scope; `_auto_deposit_for_402` checks membership first and raises the "not supported" error before the dispatch even runs. The remaining `if/elif` is now over the supported set only, with an `# pragma: no cover` else-branch that catches the "added to the set but no dispatch branch" mismatch. A sixth `PaymentType` now has to touch both the set and the dispatch — the gap surfaces at the set check instead of at offchain-402 runtime.

mech_client/infrastructure/ipfs/metadata.py
DIvyaNautiyal07: Done in 9742f52. Split out `compute_cidv1_bytes(content) -> bytes` returning the raw CID bytes (version + codec + multihash). `compute_cidv1` is now a one-liner over it. `metadata.py` does `"f" + cid_bytes.hex()` directly, dropping both the `multibase` + `multicodec` imports and the `remove_prefix` dependency. Output is identical to the previous round-trip (`f01701220<digest>`), locked in by the existing `test_full_hash_is_v1_hex` assertion.

(Side note on the suggestion's `cid_bytes[2:]`: the raw bytes start with `[0x01 (version), 0x70 (codec), 0x12, 0x20, <32 digest>]`, so `cid_bytes[2:]` skips both version AND codec — would emit `f011220...` and miss the `70`. Used `"f" + cid_bytes.hex()` instead, same result.)

mech_client/cli/commands/request_cmd.py
DIvyaNautiyal07: Done in 9742f52. `--auto-deposit` is now `is_flag=True, default=False` — bare `--auto-deposit` enables it. Also dropped the `auto_deposit = auto_deposit or False` line in the command body since the flag form is never None.

You're right that the same fix applies to `--use-prepaid` / `--use-offchain`, but those are a pre-existing convention and the example at L139 already pre-dates this PR. Left them out of scope here.

tests/unit/services/test_marketplace_service.py
DIvyaNautiyal07: Done in 9742f52. `test_log_payment_receipt_present_logs_value` now asserts the receipt value is in the log call args. Used `patch("...logger.info")` rather than `caplog` because `mech_client` sets `propagate=False` on its root logger (`mech_client/utils/logger.py:112`), so caplog records come back empty — added a comment explaining that. The absent-receipt test now also asserts `logger.info` is never called, so a refactor that always-logs would fail loudly.

tests/unit/services/test_marketplace_service.py
DIvyaNautiyal07: Done in 9742f52. Regex is now `match=r"need 100.*0xbt"` — broad enough to allow the message to be reworded around the amount and address, tight enough that dropping either from the message fails the test.

OjusWiZard (dismissed):
Re-reviewed the fix commits (`9742f52`, `ff8593d`) — all 8 review threads addressed and verified correct:

- Second 402 after auto-deposit now raises a deposit-aware error (+ regression test); duplicated POST extracted to `_do_offchain_post`.
- Malformed/non-numeric 402 bodies handled via `_safe_int` + warning log.
- Challenge is now a typed `PaymentChallenge` with a `shortfall` property.
- Auto-deposit dispatch gated by an explicit `_SUPPORTED_DEPOSIT_PAYMENT_TYPES` frozenset.
- CID hex form simplified to `compute_cidv1_bytes` (verified byte-identical `f01701220…` output, oversize guard intact, no more multibase/multicodec round-trip).
- `--auto-deposit` is now `is_flag=True`; receipt-logging and error-message tests assert real behavior.

One optional nit: the second-402 message says "deposited to …" even on the rare path where a malformed body zeroes the challenge and no deposit actually happens — edge-of-an-edge, non-blocking. The pre-existing `data=payload` vs `Content-Type: application/json` form-encoding is unchanged and out of scope.

LGTM 🚀

mech_client/services/marketplace_service.py
```diff
@@ -355,6 +421,161 @@ async def _send_offchain_request(  # pylint: disable=too-many-arguments,too-many
...
+            if response.status_code == HTTP_PAYMENT_REQUIRED:
+                # The deposit tx already confirmed (DepositService waits for
+                # receipt), so silently returning the second 402 would surface
+                # downstream as a generic "Payment Required (HTTP 402)" error
+                # with no hint that funds moved. Raise explicitly so the user
+                # sees how much is still short and which balance tracker holds
+                # the deposit they just paid for.
+                retry = self._parse_402_challenge(response)
+                raise ValueError(
+                    "Auto-deposit did not clear the 402: deposited to "
```
bennyjo, line 501: **Impact:** this error tells the user "funds were deposited" when none were. Reached whenever `_auto_deposit_for_402` returns early at line 532 (`shortfall <= 0`) — i.e. a malformed/non-numeric 402 body (`_safe_int` → 0) or `current_balance >= required` but still a 402. The message then contradicts reality precisely when someone is reading it to debug a mech regression.

Skip the retry when `shortfall <= 0`, or word the message on whether a deposit actually fired? Fine to leave if you think zero-shortfall-but-still-402 can't happen.

mech_client/services/marketplace_service.py
```diff
@@ -355,6 +421,161 @@ async def _send_offchain_request(  # pylint: disable=too-many-arguments,too-many
...
+        )
+        logger.info(
+            f"Auto-depositing {challenge.shortfall} to top up the prepaid balance..."
+        )
+        # Per-asset dispatch over the supported-types frozenset. A future
+        # payment type would need to be added to _SUPPORTED_DEPOSIT_PAYMENT_TYPES
+        # AND get a branch here; the assertion-shaped raise below catches the
+        # mismatch if only the set is updated.
+        if payment_type == PaymentType.NATIVE:
+            deposit_service.deposit_native(challenge.shortfall)
```
bennyjo, line 615: **Impact:** with `--auto-deposit`, a compromised or buggy mech (URL is auto-discovered) can return a huge `required` and drain the wallet up to its full balance into the prepaid tracker in one retry — no cap, no confirmation. Not theft (funds land in the user's own tracker, `check_balance` bounds it), but forced overspend / fund-lock.

Deliberate? A `--max-deposit` ceiling or a confirmation above some multiple of `max_delivery_rate` would close it; otherwise a one-line note on the `--auto-deposit` flag that it trusts the mech's amount. Question, not a blocker.

bennyjo (commented):
Two questions on the auto-deposit 402 path — both framed as questions, neither a blocker.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Fixed in 687ce9c. When `challenge.shortfall <= 0`, `_post_offchain_request` now raises directly without running the deposit + retry. The new message:

```
Offchain request rejected with HTTP 402 but the challenge reports no
shortfall (required=100, current balance=100). The mech may be returning a
malformed body or a stale balance. (insufficient balance)
```

It names the two likely causes (malformed body and stale balance) and never claims a deposit happened. `test_post_offchain_request_zero_shortfall_skips_retry` locks both properties down: exactly one POST happens (no retry, no deposit), and the message asserts `"deposited to" not in str(exc_info.value)`.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Closed in 687ce9c. Added `_MAX_AUTO_DEPOSIT_RATIO = 10` and plumbed `max_delivery_rate` through `_post_offchain_request` → `_auto_deposit_for_402`. When the demanded shortfall exceeds `10 * max_delivery_rate`, the deposit is refused:

```
Auto-deposit refused: mech demanded a 1001 shortfall which exceeds the safety
cap of 1000 (10x your signed max_delivery_rate of 100). If this is
legitimate, deposit the amount manually so it leaves your wallet under your
direct control.
```

10x gives slack for legitimate price moves between signature and request, tight enough to bound catastrophic drain. CLI `--auto-deposit` help text now mentions the cap. Covered by `test_auto_deposit_at_exactly_the_cap_proceeds` (at-cap allowed), `test_auto_deposit_above_the_cap_is_refused` (1-above-cap refused with the specific error), and `test_post_offchain_request_above_cap_is_refused` (asserts no retry POST and no deposit fire).

Didn't add a `--max-deposit` CLI flag — wanted the safe behaviour to be the default with no flag needed. Happy to add one if you'd rather give users the override knob.

bennyjo (approved):
Re-reviewed `687ce9c` — both findings from my prior pass are addressed correctly and covered by discriminating tests:

- **Honest error on zero-shortfall 402.** `_post_offchain_request` now guards `challenge.shortfall <= 0` before the deposit+retry, raising a message that names the situation without claiming a deposit happened. `test_post_offchain_request_zero_shortfall_skips_retry` asserts one POST, no deposit, and `"deposited to" not in` the error — exactly what locks out the lying message.
- **Auto-deposit cap at 10x max_delivery_rate.** `_MAX_AUTO_DEPOSIT_RATIO` + the `shortfall > cap` refusal bound a hostile/buggy `required`; the at-cap and above-cap boundary tests pin both sides, and the refusal propagates through the POST path with no retry.

As a bonus the new `shortfall <= 0` guard also closes the earlier optional nit about the second-402 `deposited to …` message — it's now only reachable after a deposit actually fired. The deposit destination/asset come from the client's own config + payment_type dispatch, not from mech-supplied `payTo`/`asset`, so a hostile mech can't redirect funds.

LGTM 🚀

OjusWiZard (approved):

PR #243, opened 2026-07-07
fix(agent-mode): avoid EIP-150 63/64 revert when executing Safe transactions

### LINKED ISSUE #242: Agent mode: Marketplace request() reverts (-32000) due to EIP-150 63/64 gas problem in SafeClient
## Summary

In **agent mode** (Safe-based), submitting a Mech Marketplace request fails at submission. `SafeClient.send_transaction` swallows `Exception while sending Safe transaction: {'code': -32000, 'message': 'Transaction execution fails'}`, returns `None`, and `AgentExecutor` then raises `Failed to execute Safe transaction`. The on-chain `request()` is never sent.

This hits any gas-heavy inner call (e.g. `MechMarketplace.request()`, ~340k gas) routed through a threshold-1 Safe.

## Version
- `mech-client` 0.21.1, `web3` 7.16.0, bundled `safe-eth-py`
- Chain: Gnosis (chain id 100), threshold-1 Safe (single agent EOA owner)

## Root cause

`mech_client/infrastructure/blockchain/safe_client.py::SafeClient.send_transaction` sets a **non-zero** `safe_tx_gas` from `estimate_tx_gas_with_safe(...)`:

```python
estimated_gas = self.safe.estimate_tx_gas_with_safe(to=to_address, value=value,
                                                    data=bytes.fromhex(tx_data[2:]), operation=0)
safe_tx = self.safe.build_multisig_tx(..., safe_tx_gas=estimated_gas, ...)
safe_tx.sign(signer_private_key)
tx_hash, _ = safe_tx.execute(signer_private_key)   # no explicit tx_gas
```

With a non-zero `safe_tx_gas` and no explicit `tx_gas`, `safe-eth-py`'s `SafeTx.execute()` lets web3 `eth_estimateGas` the **outer** `execTransaction`. Due to the EIP-150 "63/64" gas-forwarding rule, the node can't allocate a total gas such that 63/64 of the remaining gas still covers the inner `safe_tx_gas`, so the estimate reverts with `-32000 Transaction execution fails`. `safe-eth-py` logs the tell-tale warning:

```
Safe=0x... Found 63/64 problem gas-estimated=341978 to=<marketplace> data=0xf6938b09...
```

It is **not** a balance issue (outer gas cost is ~0 on Gnosis) and **not** a marketplace/requester issue: calling `request()` directly as `msg.sender = <Safe>` via `eth_call`, and the full `execTransaction` built with `safe_tx_gas=0`, both simulate cleanly.

## Reproduction
1. Provision an Olas service (threshold-1 Safe owned by the agent EOA), e.g. via Pearl/quickstart.
2. Agent-mode request to a live native-payment mech (e.g. Gnosis service id `2182`, mech `0xc05e7412439bd7e91730a6880e18d5d5873f632c`):
   ```python
   from mech_client.services.marketplace_service import MarketplaceService
   from safe_eth.eth import EthereumClient
   svc = MarketplaceService(chain_config="gnosis", agent_mode=True, crypto=<agent EOA>,
                            safe_address=<service Safe>, ethereum_client=EthereumClient(rpc))
   svc.send_request(prompts=("...",), tools=("superforcaster",),
                    priority_mech="0xc05e7412439bd7e91730a6880e18d5d5873f632c")
   ```
3. Fails at "Submitting marketplace request transaction" with `-32000 Transaction execution fails`.

## Suggested fix

In `SafeClient.send_transaction`, forward all gas to the inner call (`safe_tx_gas=0`) and pass an explicit outer `tx_gas` so web3 skips the failing `estimate_gas`:

```python
inner = self.safe.estimate_tx_gas_with_safe(to=to_address, value=value,
                                            data=bytes.fromhex(tx_data[2:]), operation=0)
safe_tx = self.safe.build_multisig_tx(
    to=to_address, value=value, data=bytes.fromhex(tx_data[2:]), operation=0,
    safe_tx_gas=0, base_gas=0, gas_price=0,
    gas_token=ADDRESS_ZERO, refund_receiver=ADDRESS_ZERO,
)
safe_tx.sign(signer_private_key)
tx_hash, _ = safe_tx.execute(signer_private_key, tx_gas=int(inner * 1.1) + 250_000)
```

With `safe_tx_gas=0` the inner call receives all forwarded gas, and the explicit `tx_gas` avoids the outer `estimate_gas`. Verified on Gnosis mainnet: the same request that reverts with the current code submits successfully and the mech delivers, using this approach.

Also worth surfacing the swallowed exception in `send_transaction` (it currently `logger.error`s and returns `None`), so callers get the real reason instead of a generic `Failed to execute Safe transaction`.

OjusWiZard:
## Summary

Fixes #242 — in agent mode, marketplace requests failed at submission with `-32000 Transaction execution fails` because of the EIP-150 63/64 gas-forwarding problem.

`SafeClient.send_transaction` built the multisig tx with a **non-zero** `safe_tx_gas` (the inner-call estimate) and let web3 `eth_estimateGas` the outer `execTransaction`. For gas-heavy inner calls (e.g. `MechMarketplace.request()`, ~340k gas) the node cannot find a total gas such that 63/64 of the remaining gas still covers the inner `safe_tx_gas`, so the estimate reverts and the request never reaches the chain.

## Changes

- **`SafeClient.send_transaction`** now builds the Safe tx with `safe_tx_gas=0`, so the Safe forwards all available gas to the inner call, and passes an explicit outer `tx_gas` (`inner estimate * 1.1 + 250_000` overhead, via new `OUTER_TX_GAS_MULTIPLIER` / `OUTER_TX_GAS_OVERHEAD` constants) to `SafeTx.execute()`, which makes web3 skip the failing outer `eth_estimateGas`.
- **Error surfacing**: `send_transaction` no longer swallows exceptions and returns `None` (return type `Optional[HexBytes]` → `HexBytes`). `AgentExecutor.execute_transaction` / `execute_transfer` wrap the underlying cause, so users see `Failed to execute Safe transaction: <actual reason>` instead of the bare generic message.
- Tests updated for the new contract: assert `safe_tx_gas == 0`, the explicit `tx_gas` on `execute()`, and that failure reasons propagate.

## Design note: error logging moved out of `SafeClient`

The old code logged `Exception while sending Safe transaction: ...` inside `SafeClient` and then returned `None`, so the same failure was reported twice (once in logs, once as the caller's generic exception) while the actual cause never reached the caller. With this PR, `SafeClient` is a thin infrastructure wrapper that **raises instead of logging** — the full reason travels up in the exception (`__cause__` preserved via `raise ... from e`), and the layer that owns error presentation decides how to report it. Today that is `AgentExecutor`, which wraps the cause into its `Failed to execute Safe transaction: <reason>` message that surfaces through the CLI error handlers. Downstream consumers embedding `SafeClient`/`AgentExecutor` directly should add a log/`logger.exception` at their own catch site if they swallow the exception — there is intentionally no log line inside the library layer anymore, to avoid double-reporting every failure.

## Verification

- Full unit suite: 729 passed; changed modules at 100% coverage.
- Lint: black, isort, flake8, mypy, pylint, darglint all green via `tomte tox`.
- The gas approach matches what the issue reporter verified end-to-end on Gnosis mainnet (same request that reverted submits successfully and the mech delivers), and the bundled safe-eth-py source confirms passing `tx_gas` bypasses the outer gas estimation.

### DIFF
--- a/mech_client/domain/execution/agent_executor.py
+++ b/mech_client/domain/execution/agent_executor.py
@@ -92,15 +92,15 @@ class AgentExecutor(TransactionExecutor):
 
         # Execute through Safe
         value = tx_args.get("value", 0)
-        tx_hash = self.safe_client.send_transaction(
-            to_address=contract.address,
-            tx_data=transaction["data"],
-            signer_private_key=self.private_key,
-            value=value,
-        )
-
-        if tx_hash is None:
-            raise Exception("Failed to execute Safe transaction")
+        try:
+            tx_hash = self.safe_client.send_transaction(
+                to_address=contract.address,
+                tx_data=transaction["data"],
+                signer_private_key=self.private_key,
+                value=value,
+            )
+        except Exception as e:  # pylint: disable=broad-except
+            raise Exception(f"Failed to execute Safe transaction: {e}") from e
 
         return tx_hash.to_0x_hex()
 
@@ -119,15 +119,15 @@ class AgentExecutor(TransactionExecutor):
         :return: Transaction hash
         :raises Exception: If transaction fails
         """
-        tx_hash = self.safe_client.send_transaction(
-            to_address=to_address,
-            tx_data="0x",
-            signer_private_key=self.private_key,
-            value=amount,
-        )
-
-        if tx_hash is None:
-            raise Exception("Failed to execute Safe transfer transaction")
+        try:
+            tx_hash = self.safe_client.send_transaction(
+                to_address=to_address,
+                tx_data="0x",
+                signer_private_key=self.private_key,
+                value=amount,
+            )
+        except Exception as e:  # pylint: disable=broad-except
+            raise Exception(f"Failed to execute Safe transfer transaction: {e}") from e
 
         return tx_hash.to_0x_hex()
 
--- a/mech_client/infrastructure/blockchain/safe_client.py
+++ b/mech_client/infrastructure/blockchain/safe_client.py
@@ -27,11 +27,14 @@ from safe_eth.safe import Safe  # pylint:disable=import-error
 from web3.constants import ADDRESS_ZERO
 
 # Headroom applied on top of the inner-call gas estimate when sizing the outer
-# execTransaction: the multiplier absorbs estimation variance, and the fixed
-# overhead covers Safe plumbing (signature checks, events) plus the reserve the
-# EIP-150 63/64 forwarding rule withholds from the inner call.
+# execTransaction: the multiplier absorbs estimation variance, and the
+# overhead covers Safe plumbing (signature checks, events) plus the EIP-150
+# rule capping what the inner call can receive at 63/64 of the remaining gas.
+# The overhead scales with the estimate (5%, floored at 250k) so it stays
+# proportionate for inner calls approaching the block gas limit.
 OUTER_TX_GAS_MULTIPLIER = 1.1
-OUTER_TX_GAS_OVERHEAD = 250_000
+OUTER_TX_GAS_OVERHEAD_FLOOR = 250_000
+OUTER_TX_GAS_OVERHEAD_SHARE = 0.05
 
 
 class SafeClient:
@@ -109,7 +112,11 @@ class SafeClient:
         # Sign and execute; the explicit tx_gas skips the outer
         # eth_estimateGas, which fails for the same 63/64 reason
         safe_tx.sign(signer_private_key)
-        tx_gas = int(estimated_gas * OUTER_TX_GAS_MULTIPLIER) + OUTER_TX_GAS_OVERHEAD
+        overhead = max(
+            OUTER_TX_GAS_OVERHEAD_FLOOR,
+            int(estimated_gas * OUTER_TX_GAS_OVERHEAD_SHARE),
+        )
+        tx_gas = int(estimated_gas * OUTER_TX_GAS_MULTIPLIER) + overhead
         tx_hash, _ = safe_tx.execute(signer_private_key, tx_gas=tx_gas)
         return tx_hash
 
--- a/mech_client/infrastructure/blockchain/safe_client.py
+++ b/mech_client/infrastructure/blockchain/safe_client.py
@@ -19,7 +19,6 @@
 
 """Gnosis Safe multisig client for agent mode transactions."""
 
-import logging
 from typing import Optional
 
 from hexbytes import HexBytes
@@ -27,7 +26,12 @@ from safe_eth.eth import EthereumClient  # pylint:disable=import-error
 from safe_eth.safe import Safe  # pylint:disable=import-error
 from web3.constants import ADDRESS_ZERO
 
-logger = logging.getLogger(__name__)
+# Headroom applied on top of the inner-call gas estimate when sizing the outer
+# execTransaction: the multiplier absorbs estimation variance, and the fixed
+# overhead covers Safe plumbing (signature checks, events) plus the reserve the
+# EIP-150 63/64 forwarding rule withholds from the inner call.
+OUTER_TX_GAS_MULTIPLIER = 1.1
+OUTER_TX_GAS_OVERHEAD = 250_000
 
 
 class SafeClient:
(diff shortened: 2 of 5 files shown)


DIvyaNautiyal07 (dismissed):

mech_client/infrastructure/blockchain/safe_client.py
```diff
@@ -19,15 +19,19 @@
...
 from hexbytes import HexBytes
 from safe_eth.eth import EthereumClient  # pylint:disable=import-error
 from safe_eth.safe import Safe  # pylint:disable=import-error
 from web3.constants import ADDRESS_ZERO
 
-logger = logging.getLogger(__name__)
+# Headroom applied on top of the inner-call gas estimate when sizing the outer
+# execTransaction: the multiplier absorbs estimation variance, and the fixed
+# overhead covers Safe plumbing (signature checks, events) plus the reserve the
+# EIP-150 63/64 forwarding rule withholds from the inner call.
```
jmoreira-valory: **NIT (wording) + LOW (sizing):**

**Wording:** EIP-150 doesn't _withhold_ from the inner call — it caps the max gas the inner call can _receive_ at `63/64 × gas_remaining`. "Reserve" ≈ "1/64 leftover for the outer frame", if that framing helps.

**Sizing:** `OUTER_TX_GAS_OVERHEAD = 250_000` is one-size-fits-all across all inner-call sizes. For a 30k inner (e.g. a token transfer), the added 250k overhead is ~800% of what's needed — the user pre-authorises a gas budget significantly larger than actual usage (minor wallet-UX issue on some frontends; actual on-chain cost is only `gas_used` so no fee impact). For inner calls approaching the block limit, it can be tight. `overhead = max(250_000, int(estimated_gas * 0.05))` fits both. Non-blocking — the marketplace-request path this fixes is comfortably in range.

mech_client/infrastructure/blockchain/safe_client.py
```diff
@@ -66,46 +70,48 @@ def send_transaction(  # pylint: disable=too-many-arguments
         tx_data: str,
         signer_private_key: str,
         value: int = 0,
-    ) -> Optional[HexBytes]:
+    ) -> HexBytes:
         """
         Build, sign, and execute a Safe multisig transaction.
 
+        Exceptions from building, signing, or executing the transaction
+        propagate to the caller instead of being swallowed.
```
jmoreira-valory, line 81: **LOW — observability regression.** The old code emitted `logger.error(f"Exception while sending Safe transaction: {e}")` at the point of failure before returning `None`. The new contract propagates exceptions instead (good), but `logger` was removed entirely from this module — no log line fires from here anymore. Any caller that doesn't add their own `logger.exception("Safe transaction failed")` before re-catching (or lets it bubble to a bare `except Exception`) will lose the failure in logs. Reasonable for a library — callers own their logging — but worth calling out the surface change so downstream consumers know to add a log at their layer.

mech_client/domain/execution/agent_executor.py
```diff
@@ -92,15 +92,15 @@ def execute_transaction(
...
-            raise Exception("Failed to execute Safe transaction")
+        try:
+            tx_hash = self.safe_client.send_transaction(
+                to_address=contract.address,
+                tx_data=transaction["data"],
+                signer_private_key=self.private_key,
+                value=value,
+            )
+        except Exception as e:  # pylint: disable=broad-except
+            raise Exception(f"Failed to execute Safe transaction: {e}") from e
```
jmoreira-valory, line 103: **NIT — exception type flattening.** `except Exception as e: raise Exception(f"...") from e` flattens the specific type (e.g. `web3.exceptions.ContractLogicError`, `ValueError`) to bare `Exception`. Downstream callers catching by type can't anymore. `raise … from e` preserves `__cause__` so introspection still works via `.__cause__`, but a caller that does `except ContractLogicError` will now miss it. Consistent with the existing raise-generic-`Exception` pattern in this codebase, so probably intentional — flagging in case narrower types were preferable.

tox.ini
```diff
@@ -145,8 +145,9 @@ trio: ==0.32.0
 zipp: ==3.23.0
 ; MIT
 argon2-cffi-bindings: ==25.1.0
-; MIT
-cffi: 2.0.0
+; MIT (2.0.0 ships no license metadata; 2.1.0+ declares MIT-0, which is
```
jmoreira-valory: **NIT — comment wording.** The comment reads like it's describing a problem the change _creates_ ("2.1.0+ declares MIT-0, which is not in tomte's authorized_licenses list"). What it's actually saying is _why_ the entry is here in `[Authorized Packages]`: listing `cffi: >=2.0.0` bypasses liccheck's license check entirely, because neither 2.0.0 (no metadata) nor 2.1.0+ (MIT-0) would pass a strict license lookup. One extra sentence — e.g. `; listing here bypasses liccheck's license check for all >=2.0.0 releases` — makes the intent explicit for the next reader.

jmoreira-valory (dismissed):
Independent Mode-A review. Well-scoped `safe_tx_gas=0` + explicit outer `tx_gas` fix — matches the safe-eth-py pattern for bypassing EIP-150. Sanity-checked the math on a ~340k inner call: outer `tx_gas=624k`, Safe forwards `~574k` to the inner (after `execTransaction` boilerplate + 63/64 cap), plenty of headroom. Tests updated cleanly, 100% cov on changed modules, CI 30/30 green.

Posting a few non-blocking notes inline. LGTM.

mech_client/infrastructure/blockchain/safe_client.py
OjusWiZard: Good catch — this is intentional, and we've decided to keep the library layer log-free rather than reintroduce a logger. Rationale (now also spelled out in the PR description under "Design note: error logging moved out of `SafeClient`"): the old log-then-return-`None` pattern reported every failure twice while hiding the actual cause from the caller. Now the full reason travels up in the exception itself (`__cause__` preserved via `raise ... from e`), and the layer that owns error presentation — `AgentExecutor` → CLI error handlers — reports it once, with the real cause embedded in the message. Downstream consumers embedding `SafeClient`/`AgentExecutor` directly should log at their own catch site if they swallow the exception.

DIvyaNautiyal07 (approved):

jmoreira-valory (approved):
Recheck at HEAD `9e1274480b0bc7bfa8d69edfaf8de40ffcaf0bcb`. All three actionable items from my prior review addressed cleanly:

- **Sizing (LOW):** `OUTER_TX_GAS_OVERHEAD_FLOOR` + `OUTER_TX_GAS_OVERHEAD_SHARE`, computed as `max(250_000, int(estimated_gas * 0.05))` — matches my suggested formula. Regression test `test_send_transaction_large_inner_call_scales_overhead` verifies the proportional-share branch.
- **Wording (NIT):** header comment now reads "the EIP-150 rule capping what the inner call can receive at 63/64 of the remaining gas" — accurate framing.
- **tox.ini (NIT):** intent explicit — "Listing here bypasses liccheck's license check for all >=2.0.0 releases".

LOW observability regression (removed `logger.error`) and NIT exception type flattening in `AgentExecutor` are consistent with library-caller-owns-logging semantics. Not blocking. CI 30/30 green. LGTM (re-approve after dismissal).

PR #245, opened 2026-07-07
feat(signing): support external signers via an injectable Signer interface

### LINKED ISSUE #244: Support external signers: route all signing through an injectable Signer interface (no raw private key required)
## Motivation

Some mech-client consumers cannot hold a raw private key in-process. Example: Pearl BYOA-style agents, where key custody lives in a separate local signer service and the code driving mech-client only gets `sign`/`send` capabilities over an API — never the key bytes.

Today every signing path in mech-client requires the raw key (a `crypto` object / `private_key`), which makes both the on-chain agent-mode flow and the off-chain prepaid flow unusable in this setup.

## Proposal

Introduce a signer abstraction and route all signing through it, keeping the current behavior as the default implementation.

```python
class Signer(Protocol):
    address: str

    def send_transaction(self, unsigned_tx: dict) -> str:
        """Sign and broadcast an EOA transaction; return the tx hash.
        unsigned_tx: {chainId, to, value, data, ...} — implementation fills nonce/gas if absent."""

    def sign_message(self, message: bytes) -> bytes:
        """Return an EIP-191 signature over `message`."""
```

- Constructors of `MarketplaceService` / `DepositService` accept `signer=` as an alternative to `crypto=`.
- Default implementation (`LocalSigner`, wrapping the private key as today) preserves full backwards compatibility.

## Call sites that must go through the signer

1. **Agent-mode SafeTx execution** (`MarketplaceService`, via safe-eth-py):
   - Replace `SafeTx.sign(private_key)` with the **pre-validated signature** encoding (`r` = owner address left-padded to 32 bytes, `s = 0`, `v = 1`). Valid whenever the outer `execTransaction` sender is that owner — no ECDSA over the SafeTx hash needed at all.
   - Replace `SafeTx.execute(private_key, ...)` with building the outer `execTransaction` transaction (e.g. `SafeTx.w3_tx` / `build_transaction`) and submitting it via `signer.send_transaction(...)`.
2. **Deposits** (`DepositService.deposit_native()` and token variants): plain EOA transactions — build unsigned, submit via `signer.send_transaction(...)`.
3. **Off-chain requests** (`MarketplaceService.send_request(use_offchain=True)`): the EIP-191 request signature — obtain via `signer.sign_message(...)`.

## Acceptance criteria

With only a `Signer` implementation (no private key anywhere in the mech-client process), all of the following work end to end:

- [ ] on-chain marketplace request paid by a safe (agent mode), executed via pre-validated signature + externally signed outer tx
- [ ] `deposit_native()` (and token deposit) to the BalanceTracker
- [ ] off-chain prepaid request (`use_prepaid=True, use_offchain=True`)
- [ ] existing `crypto=`/private-key usage unchanged (LocalSigner default)

## Out of scope

The `safe_tx_gas` / EIP-150 gas problem in agent-mode `SafeClient` is tracked separately in #242 and is not part of this issue.

OjusWiZard:
## Summary

Closes #244.

Some mech-client consumers cannot hold a raw private key in-process (e.g. Pearl BYOA-style agents, where key custody lives in a separate local signer service). This PR introduces a `Signer` abstraction and routes **all** signing through it, with the existing behavior preserved as the default.

```python
@runtime_checkable
class Signer(Protocol):
    @property
    def address(self) -> str: ...
    def send_transaction(self, unsigned_tx: dict) -> str: ...
    def sign_message(self, message: bytes) -> bytes: ...
```

- `MarketplaceService` / `DepositService` (and the shared `BaseTransactionService`) accept `signer=` as an alternative to `crypto=`.
- `LocalSigner` (wrapping `EthereumCrypto` + ledger API) is the default and preserves historic signing behavior byte-for-byte.

## Call sites routed through the signer

1. **Agent-mode SafeTx execution** — `SafeTx.sign(private_key)` / `SafeTx.execute(private_key)` are replaced by the Safe **pre-validated signature** encoding (`r` = owner address left-padded to 32 bytes, `s = 0`, `v = 1`), valid because the outer `execTransaction` sender is that owner. The outer tx is built via `SafeTx.w3_tx` and submitted through `signer.send_transaction(...)`. Outer gas sizing follows #243 (merged separately on `main`): `safe_tx_gas=0` forwards all gas to the inner call, and an explicit `tx_gas = int(estimate * 1.1) + max(250k, 5% of estimate)` skips the outer `eth_estimateGas` that the EIP-150 63/64 rule would otherwise revert — this PR only changes *who signs and broadcasts* that outer tx, not how its gas is sized. No ECDSA over the SafeTx hash is needed, so no raw-hash signing capability is required from the signer service.
2. **Deposits** — native and token deposits build unsigned EOA transactions and submit via `signer.send_transaction(...)` (client mode) or the Safe path above (agent mode).
3. **Off-chain requests** — the request-id signature comes from `signer.sign_message(...)`. Note: the on-chain verifier (`MechMarketplace.sol`) recovers with plain `ecrecover(requestHash, v, r, s)` on the **raw 32-byte digest** — no EIP-191 prefix (the issue text says "EIP-191", but the contract and the historic `unsafe_sign_hash` behavior both use direct digest signing; the `Signer.sign_message` docstring documents this precisely so external implementers get it right).

## Backwards compatibility

- `crypto=` keeps working unchanged everywhere (wrapped in `LocalSigner` internally); `SubscriptionService` and the CLI are untouched.
- Offchain signature bytes and payload format are byte-for-byte identical to before.
- `SafeClient.send_transaction` returns `HexBytes` and propagates failures as exceptions (matching post-#243 `main`); malformed signer addresses fail with a diagnostic `ValueError`.
- Per review feedback, services no longer store `crypto` / `private_key` attributes (they were dead state — all signing goes through `service.signer`); the `crypto=` constructor argument itself is unchanged.

## Acceptance criteria (from #244)

- [x] Agent-mode marketplace request paid by a Safe, executed via pre-validated signature + externally signed outer tx
- [x] `deposit_native()` and token deposits through the signer
- [x] Off-chain prepaid request signing through the signer
- [x] Existing `crypto=`/private-key usage unchanged (`LocalSigner` default)

## Docs

- `docs/ARCHITECTURE.md`: new `domain/signing/` section
- `README.md`: "Using an external signer (no private key in-process)" guide
- `mech_client/__init__.py`: exports `Signer` / `LocalSigner` with a usage example

## Testing

- 747 unit tests pass, coverage stays at 100% (2867/2867 statements)
- New `tests/unit/domain/test_signing.py` (protocol conformance, `LocalSigner` nonce/gas fill, EIP-1559 non-override, input immutability, digest signing)
- Safe client tests rewritten for the pre-validated flow (encoding layout, gas floor, owner-address validation, failure paths); executor and service tests updated; signer-only service construction covered
- All linters green: black, isort, flake8, mypy, pylint (10.00/10), darglint, bandit, copyright check

### DIFF
--- a/mech_client/domain/signing/local.py
+++ b/mech_client/domain/signing/local.py
@@ -66,7 +66,7 @@ class LocalSigner:
             tx["nonce"] = self.ledger_api.api.eth.get_transaction_count(self.address)
         if "gasPrice" not in tx and "maxFeePerGas" not in tx:
             tx["gasPrice"] = self.ledger_api.api.eth.gas_price
-        if not tx.get("gas"):
+        if tx.get("gas") is None:
             tx["gas"] = self.ledger_api.api.eth.estimate_gas(tx)
         signed_transaction = self.crypto.sign_transaction(tx)
         transaction_digest = self.ledger_api.send_signed_transaction(
--- a/mech_client/domain/signing/local.py
+++ b/mech_client/domain/signing/local.py
@@ -1,0 +1,89 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2026 Valory AG
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""Local signer wrapping an in-process private key (default implementation)."""
+
+from typing import Any, Dict
+
+from aea_ledger_ethereum import EthereumApi, EthereumCrypto
+
+
+class LocalSigner:
+    """Default :class:`Signer` implementation backed by a local private key.
+
+    Wraps an ``EthereumCrypto`` (raw key in-process) and an ``EthereumApi``
+    (for nonce/gas fill and broadcasting), preserving the historic signing
+    behavior of mech-client.
+    """
+
+    def __init__(self, crypto: EthereumCrypto, ledger_api: EthereumApi):
+        """
+        Initialize local signer.
+
+        :param crypto: Ethereum crypto object holding the private key
+        :param ledger_api: Ethereum API used to fill tx defaults and broadcast
+        """
+        self.crypto = crypto
+        self.ledger_api = ledger_api
+
+    @property
+    def address(self) -> str:
+        """The EOA address of the wrapped key.
+
+        :return: Signer's EOA address
+        """
+        return str(self.crypto.address)
+
+    def send_transaction(self, unsigned_tx: Dict[str, Any]) -> str:
+        """Sign an EOA transaction with the local key and broadcast it.
+
+        Fills ``nonce``, gas price, and ``gas`` from the connected node when
+        absent, then signs and sends via the ledger API (``raise_on_try=True``
+        propagates build/sign/send failures).
+
+        :param unsigned_tx: Unsigned transaction dict
+        :return: Transaction hash (0x-prefixed hex string)
+        """
+        tx = dict(unsigned_tx)
+        tx.setdefault("from", self.address)
+        if tx.get("nonce") is None:
+            tx["nonce"] = self.ledger_api.api.eth.get_transaction_count(self.address)
+        if "gasPrice" not in tx and "maxFeePerGas" not in tx:
+            tx["gasPrice"] = self.ledger_api.api.eth.gas_price
+        if not tx.get("gas"):
+            tx["gas"] = self.ledger_api.api.eth.estimate_gas(tx)
+        signed_transaction = self.crypto.sign_transaction(tx)
+        transaction_digest = self.ledger_api.send_signed_transaction(
+            signed_transaction,
+            raise_on_try=True,
+        )
+        return str(transaction_digest)
+
+    def sign_message(self, message: bytes) -> bytes:
+        """Sign ``message`` with the local key.
+
+        For 32-byte inputs (the mech request-id digest) this signs the digest
+        directly (``unsafe_sign_hash``), matching the marketplace contract's
+        plain ``ecrecover`` verification — see :class:`Signer.sign_message`.
+
+        :param message: Message bytes to sign
+        :return: 65-byte signature (r ‖ s ‖ v)
+        """
+        signature = self.crypto.sign_message(message, is_deprecated_mode=True)
+        return bytes.fromhex(signature.removeprefix("0x"))
--- a/mech_client/infrastructure/blockchain/safe_client.py
+++ b/mech_client/infrastructure/blockchain/safe_client.py
@@ -22,6 +22,7 @@
 from typing import Optional
 
 from hexbytes import HexBytes
+from mech_client.domain.signing import Signer
 from safe_eth.eth import EthereumClient  # pylint:disable=import-error
 from safe_eth.safe import Safe  # pylint:disable=import-error
 from web3.constants import ADDRESS_ZERO
@@ -67,22 +68,54 @@ class SafeClient:
             )
         return self._safe
 
-    def send_transaction(  # pylint: disable=too-many-arguments
+    @staticmethod
+    def build_pre_validated_signature(owner: str) -> bytes:
+        """
+        Encode a pre-validated Safe owner signature.
+
+        Format (Safe contracts): ``r`` = owner address left-padded to 32
+        bytes, ``s`` = 0, ``v`` = 1. The Safe accepts it whenever the outer
+        ``execTransaction`` sender is that owner — no ECDSA over the SafeTx
+        hash is needed, so the private key never has to be in-process.
+
+        :param owner: Safe owner address (0x-prefixed) sending the outer tx
+        :return: 65-byte pre-validated signature
+        :raises ValueError: If ``owner`` is not a valid 20-byte hex address
+        """
+        try:
+            owner_bytes = bytes.fromhex(owner.removeprefix("0x"))
+        except ValueError as e:
+            raise ValueError(
+                f"Invalid Safe owner address from signer: {owner!r} (not hex)"
+            ) from e
+        if len(owner_bytes) != 20:
+            raise ValueError(
+                f"Invalid Safe owner address from signer: {owner!r} "
+                f"(expected 20 bytes, got {len(owner_bytes)})"
+            )
+        return b"\x00" * 12 + owner_bytes + b"\x00" * 32 + b"\x01"
+
+    def send_transaction(
         self,
         to_address: str,
         tx_data: str,
-        signer_private_key: str,
+        signer: Signer,
         value: int = 0,
     ) -> HexBytes:
         """
-        Build, sign, and execute a Safe multisig transaction.
+        Build and execute a Safe multisig transaction via the signer.
+
+        The SafeTx carries a pre-validated signature for the signer's address
+        (which must be a Safe owner), and the outer ``execTransaction`` is
+        built unsigned and submitted through ``signer.send_transaction`` —
+        signing happens wherever the signer keeps its key.
 
         Exceptions from building, signing, or executing the transaction
         propagate to the caller instead of being swallowed.
 
         :param to_address: Destination contract/address
         :param tx_data: Transaction data (hex string starting with 0x)
-        :param signer_private_key: Private key for signing
+        :param signer: Signer for the Safe owner EOA sending the outer tx
         :param value: ETH/native token value to send (in wei)
         :return: Transaction hash
         """
@@ -109,16 +142,29 @@ class SafeClient:
             refund_receiver=ADDRESS_ZERO,
         )
 
-        # Sign and execute; the explicit tx_gas skips the outer
-        # eth_estimateGas, which fails for the same 63/64 reason
-        safe_tx.sign(signer_private_key)
+        # Pre-validated signature: valid because the outer tx sender is
+        # the owner encoded in it.
+        safe_tx.signatures = self.build_pre_validated_signature(signer.address)
+
+        # Size the outer gas from the inner estimate; the explicit gas limit
+        # skips the outer eth_estimateGas, which fails for gas-heavy inner
+        # calls for the same 63/64 reason. The nonce is left for the signer
+        # to fill.
         overhead = max(
             OUTER_TX_GAS_OVERHEAD_FLOOR,
             int(estimated_gas * OUTER_TX_GAS_OVERHEAD_SHARE),
         )
         tx_gas = int(estimated_gas * OUTER_TX_GAS_MULTIPLIER) + overhead
-        tx_hash, _ = safe_tx.execute(signer_private_key, tx_gas=tx_gas)
-        return tx_hash
+        outer_tx = safe_tx.w3_tx.build_transaction(
+            {
+                "from": signer.address,
+                "gas": tx_gas,
+                "gasPrice": self.ethereum_client.w3.eth.gas_price,
+            }
+        )
+
+        tx_hash = signer.send_transaction(outer_tx)
+        return HexBytes(tx_hash)
 
     def get_nonce(self) -> int:
         """
--- a/mech_client/services/base_service.py
+++ b/mech_client/services/base_service.py
@@ -72,8 +72,6 @@ class BaseTransactionService:  # pylint: disable=too-few-public-methods,too-many
 
         self.chain_config = chain_config
         self.agent_mode = agent_mode
-        self.crypto = crypto
-        self.private_key = crypto.private_key if crypto is not None else None
         self.safe_address = safe_address
         self.ethereum_client = ethereum_client
 
--- a/mech_client/services/base_service.py
+++ b/mech_client/services/base_service.py
@@ -24,6 +24,7 @@ from typing import Optional
 
 from aea_ledger_ethereum import EthereumApi, EthereumCrypto
 from mech_client.domain.execution import ExecutorFactory, TransactionExecutor
+from mech_client.domain.signing import LocalSigner, Signer
 from mech_client.infrastructure.config import MechConfig, get_mech_config
 from safe_eth.eth import EthereumClient
 
@@ -36,7 +37,11 @@ class BaseTransactionService:  # pylint: disable=too-few-public-methods,too-many
     - Chain configuration and mech config loading
     - Ethereum API creation from ledger config
     - Transaction executor setup (agent mode or client mode)
-    - Crypto and address management
+    - Signer and address management
+
+    Signing is routed through a :class:`Signer`. Pass ``signer=`` to keep the
+    private key out of the mech-client process (external signer service), or
+    ``crypto=`` to sign locally as before.
 
     Subclasses should call super().__init__() first, then add their own
     specific initialization.
@@ -46,23 +51,29 @@ class BaseTransactionService:  # pylint: disable=too-few-public-methods,too-many
         self,
         chain_config: str,
         agent_mode: bool,
-        crypto: EthereumCrypto,
+        crypto: Optional[EthereumCrypto] = None,
         safe_address: Optional[str] = None,
         ethereum_client: Optional[EthereumClient] = None,
+        signer: Optional[Signer] = None,
     ):
         """
         Initialize base transaction service.
 
         :param chain_config: Chain configuration name (gnosis, base, polygon, optimism)
         :param agent_mode: True for agent mode (Safe), False for client mode (EOA)
-        :param crypto: Ethereum crypto object for signing
+        :param crypto: Ethereum crypto object for local signing (alternative to signer)
         :param safe_address: Safe address (required for agent mode)
         :param ethereum_client: Ethereum client (required for agent mode)
+        :param signer: Signer for externalized signing (alternative to crypto)
+        :raises ValueError: If neither crypto nor signer is provided
         """
+        if crypto is None and signer is None:
+            raise ValueError("Either a crypto object or a signer is required")
+
         self.chain_config = chain_config
         self.agent_mode = agent_mode
         self.crypto = crypto
-        self.private_key = crypto.private_key
+        self.private_key = crypto.private_key if crypto is not None else None
         self.safe_address = safe_address
         self.ethereum_client = ethereum_client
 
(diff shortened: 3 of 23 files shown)


mech_client/services/base_service.py
```diff
@@ -46,23 +51,29 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        :raises ValueError: If neither crypto nor signer is provided
         """
+        if crypto is None and signer is None:
+            raise ValueError("Either a crypto object or a signer is required")
+
         self.chain_config = chain_config
         self.agent_mode = agent_mode
         self.crypto = crypto
-        self.private_key = crypto.private_key
+        self.private_key = crypto.private_key if crypto is not None else None
```
DIvyaNautiyal07: P1 (Dead state): `self.crypto` and `self.private_key` are stored here but never dereferenced anywhere on this branch (`git grep 'self.crypto.'` returns 0 hits post-PR — the two `self.crypto` references are pass-throughs at marketplace_service.py:238 and :639, both accepting `Optional[EthereumCrypto]`). Keeping these fields alive means every future author has to reason about the invariant "self.crypto may be None while self.signer is always non-None" for no gain. Consider dropping both fields; the two downstream `crypto=` pass-throughs then also become obsolete.

mech_client/services/marketplace_service.py
```diff
@@ -636,6 +639,7 @@ def _auto_deposit_for_402(
             crypto=self.crypto,
```
DIvyaNautiyal07: P1 (Redundant passthrough): `crypto=self.crypto` alongside `signer=self.signer`. `DepositService` -> `BaseTransactionService.__init__` picks `signer` when non-None, so `crypto=` is ignored for signing. On the external-signer path `self.crypto` is `None` and this line just persists a `None` reference in the child service. Drop `crypto=self.crypto`; `signer=self.signer` is sufficient. Same pattern at line 238 (see top-level note).

tests/unit/infrastructure/test_safe_client.py
```diff
@@ -86,18 +86,82 @@ def test_safe_property_cached(self, mock_safe_class: MagicMock) -> None:
         assert safe1 == safe2 == mock_safe_instance
 
 
+def create_mock_signer(address: str = "0x" + "ab" * 20) -> MagicMock:
```
DIvyaNautiyal07: P1 (Duplicated helper + repeated setup): `create_mock_signer` is defined identically here and in `tests/unit/domain/test_execution_strategies.py:32` (different default addresses, same body). Additionally, `tests/unit/services/test_marketplace_service.py` repeats the 3-line inline block
```python
service.signer = MagicMock()
service.signer.address = "0x" + "a" * 40
service.signer.sign_message.return_value = b"\xab" * 65
```
4x (lines 1431, 1510, 1581, 1656). Per the "extract shared stubs into reusable helpers" rule, move `create_mock_signer` to a shared conftest/helpers module and use it in all three files (or a fixture in the marketplace test).

tests/unit/domain/test_signing.py
```diff
@@ -0,0 +1,183 @@
...
+
+        signer.send_transaction(unsigned_tx)
+
+        assert unsigned_tx == {"to": "0x" + "b" * 40, "gas": 21000, "gasPrice": 1}
+
+
+class TestLocalSignerSignMessage:
+    """Tests for LocalSigner.sign_message."""
+
+    def test_signs_digest_in_deprecated_mode(self) -> None:
```
DIvyaNautiyal07, line 182: P1 (Missing crypto-level regression coverage): `crypto.sign_message` is mocked and the test only asserts that `is_deprecated_mode=True` was passed. If the AEA plugin ever changes what `is_deprecated_mode` means (or removes the flag), all mocked tests keep passing while the on-chain `ecrecover` silently rejects offchain mech requests. Add one integration-style test that constructs a real `EthereumCrypto` from a fixed key, signs a known 32-byte digest, and asserts `eth_account.Account.recover(digest, sig)` yields the expected address. That's the only test that would catch an AEA-side regression on the offchain signing path.

mech_client/services/marketplace_service.py
```diff
@@ -636,6 +639,7 @@ def _auto_deposit_for_402(
...
             safe_address=self.safe_address,
             ethereum_client=self.ethereum_client,
+            signer=self.signer,
```
DIvyaNautiyal07, line 693: P2 (Coverage gap for this line): none of the `test_auto_deposit_*` tests assert that `signer=self.signer` here is forwarded to the `DepositService` constructor — they only assert on `deposit_native.assert_called_once_with(...)`. If a future refactor drops the `signer=` kwarg on this line, no test catches it. One `mock_ds.assert_called_once_with(..., signer=service.signer, ...)` in a single test closes the gap.

mech_client/domain/signing/local.py
```diff
@@ -0,0 +1,89 @@
...
+        :param unsigned_tx: Unsigned transaction dict
+        :return: Transaction hash (0x-prefixed hex string)
+        """
+        tx = dict(unsigned_tx)
+        tx.setdefault("from", self.address)
+        if tx.get("nonce") is None:
+            tx["nonce"] = self.ledger_api.api.eth.get_transaction_count(self.address)
+        if "gasPrice" not in tx and "maxFeePerGas" not in tx:
+            tx["gasPrice"] = self.ledger_api.api.eth.gas_price
+        if not tx.get("gas"):
```
DIvyaNautiyal07: P2 (Falsy-not-None guard): `if not tx.get("gas")` triggers on `gas=0` and re-runs `estimate_gas`. Line 65 above uses the consistent form `if tx.get("nonce") is None`. A caller that intentionally passes `gas=0` — rare, but the intent is "use my value" — silently gets it overwritten. Fix: `if tx.get("gas") is None:` for consistency.

mech_client/infrastructure/blockchain/safe_client.py
```diff
@@ -109,16 +142,29 @@ def send_transaction(  # pylint: disable=too-many-arguments
...
-        # eth_estimateGas, which fails for the same 63/64 reason
-        safe_tx.sign(signer_private_key)
+        # Pre-validated signature: valid because the outer tx sender is
+        # the owner encoded in it.
+        safe_tx.signatures = self.build_pre_validated_signature(signer.address)
+
+        # Size the outer gas from the inner estimate; the explicit gas limit
+        # skips the outer eth_estimateGas, which fails for gas-heavy inner
+        # calls for the same 63/64 reason. The nonce is left for the signer
+        # to fill.
```
DIvyaNautiyal07, line 152: P2 (Inaccurate comment): "The nonce is left for the signer to fill" isn't right. `safe_tx.w3_tx.build_transaction({"from": signer.address, ...})` calls web3's `fill_transaction_defaults`, which fills `nonce` from `get_transaction_count(from)` when it's absent from the overrides dict. `outer_tx` already has a nonce by the time the signer sees it. Reword to something like: "web3's `build_transaction` fills `nonce` from `get_transaction_count(from)`; an external signer can override if it manages its own nonce sequencing."

DIvyaNautiyal07 (commented):
Two additional notes that don't map cleanly to inline positions:

**P2 — PR description doesn't call out `safe_tx_gas=estimated_gas -> 0`.** The PR body describes the pre-validated signature switch but not the `safe_tx_gas=0` change in `safe_client.py`'s `build_multisig_tx` call. The in-code comment ("safe_tx_gas=0 forwards all available gas to the inner call") explains the rationale and it's the right call, but it affects the backward-compat `LocalSigner` path too, not just external-signer callers. Worth a bullet in the description.

**P1 (see the base_service comment) — same redundancy also at `marketplace_service.py:238`.** The on-chain flow passes `crypto=self.crypto` to `PaymentStrategyFactory.create`. `TokenPaymentStrategy` stores `self.crypto` (line 72) but never dereferences it — the actual approve tx goes through `executor.execute_transaction`, not `self.crypto`. Same fix pattern as the `_auto_deposit_for_402` case flagged inline.

mech_client/infrastructure/blockchain/safe_client.py
OjusWiZard: Checked this against the installed web3 (7.15.0) before changing it — web3's `build_transaction` does **not** fill the nonce. `web3._utils.transactions.TRANSACTION_DEFAULTS` is:

```python
{'value', 'data', 'gas', 'gasPrice', 'maxPriorityFeePerGas', 'maxFeePerGas', 'chainId'}
```

`nonce` isn't in the set, and `fill_transaction_defaults` never touches it. That's also why safe-eth-py's own `SafeTx.execute()` doesn't rely on `build_transaction` for the nonce — it broadcasts via `ethereum_client.send_unsigned_transaction`, which fills the nonce itself.

So `outer_tx` genuinely leaves the signer to supply the nonce: `LocalSigner` fills it from `get_transaction_count`, and an external signer service fills it from its own sequencing. Keeping the comment as-is.

All other findings from this review are addressed in a9e270c (dead `crypto`/`private_key` state dropped, redundant `crypto=` pass-throughs and the unused `TokenPaymentStrategy` param removed, shared `create_mock_signer` helper, real-key ecrecover regression test, `signer=` forwarding assertion, `gas is None` check) — plus the PR description now covers the outer-gas sizing inherited from #243.

DIvyaNautiyal07, line 152: Small follow-up on the previous review: this comment still reads "The nonce is left for the signer to fill," but `safe_tx.w3_tx.build_transaction({"from": signer.address, ...})` calls web3's `fill_transaction_defaults`, which fills `nonce` from `get_transaction_count(from)` when the overrides don't include one. So `outer_tx` already has a nonce by the time the signer sees it — `LocalSigner`'s `if tx.get("nonce") is None` guard is defensive but doesn't fire on this path. Suggested reword: "web3's `build_transaction` fills `nonce` from `get_transaction_count(from)`; an external signer can override if it manages its own nonce sequencing."

DIvyaNautiyal07 (approved):
Follow-up commit `a9e270c` addresses everything from the earlier review cleanly. Quick summary of what landed:

**P1s — resolved**
- Dead-weight `self.crypto` / `self.private_key` on `BaseTransactionService` removed; `crypto=` param dropped from `PaymentStrategyFactory.create` and `TokenPaymentStrategy.__init__` (whole `__init__` deleted); the three `crypto=self.crypto` passthroughs are gone.
- `create_mock_signer` and the shared constants moved to `tests/unit/helpers.py`; the 4 repeated inline setup blocks in `test_marketplace_service.py` collapsed to `create_mock_signer(...)` calls.
- New `test_real_key_signature_recovers_via_raw_ecrecover` constructs a real `EthereumCrypto`, signs a 32-byte digest, and recovers with `Account._recover_hash` — the exact primitive the on-chain `ecrecover(digest, v, r, s)` mirrors. This is the test that will fail if `is_deprecated_mode=True` ever stops meaning "no EIP-191 prefix," which was the whole point.

**P2s — resolved**
- `test_auto_deposit_native` now asserts on the full `DepositService(...)` constructor kwargs including `signer=service.signer`.
- `LocalSigner.send_transaction` gas guard changed to `is None`; `test_preserves_explicit_zero_gas` locks the behavior in.

**One P2 still open** (inline note below): the stale "nonce is left for the signer to fill" comment in `safe_client.py`.

**P2 previously flagged in the top-level body** (PR description doesn't call out the `safe_tx_gas=estimated_gas -> 0` change) is still worth a bullet in the description, but that's a doc edit, not a code change.

LGTM otherwise — approving.

mech_client/infrastructure/blockchain/safe_client.py
OjusWiZard: Re-checked against the installed web3 (7.15.0) source rather than memory — `fill_transaction_defaults` cannot fill a nonce because `TRANSACTION_DEFAULTS` has no `nonce` entry (`web3/_utils/transactions.py:77-90`):

```python
TRANSACTION_DEFAULTS = {
    "value": 0,
    "data": b"",
    "gas": lambda w3, tx, _defaults: w3.eth.estimate_gas(tx),
    "gasPrice": lambda w3, tx, _defaults: w3.eth.generate_gas_price(tx),
    "maxPriorityFeePerGas": lambda w3, _tx, _defaults: w3.eth.max_priority_fee,
    "maxFeePerGas": (...),
    "chainId": lambda w3, _tx, _defaults: w3.eth.chain_id,
}
```

and `fill_transaction_defaults` only iterates `TRANSACTION_DEFAULTS.items()` — there is no other code path in `build_transaction` that touches `nonce`. So `outer_tx` reaches the signer without one, and `LocalSigner`'s `if tx.get("nonce") is None` guard does fire on this path (it's what supplies the nonce for the local-key case). Keeping the comment as-is.

DIvyaNautiyal07 (dismissed):
Re-reviewed after your follow-up. On the `safe_client.py:152` nonce comment you're right and I was wrong. Verified against the installed web3 7.15.0 source, `TRANSACTION_DEFAULTS` at `web3/_utils/transactions.py:77-90` doesn't include `nonce`, and `fill_transaction_defaults` at line 114 only iterates that dict. So `build_transaction` never fills the nonce, `outer_tx` reaches `LocalSigner` without one, and the existing comment is accurate. Apologies for doubling down on that one before checking the source. My follow-up comment 3538548618 can be resolved.

Everything else is addressed in `a9e270c`. `self.crypto` and `self.private_key` are gone from `base_service.py`, both `crypto=self.crypto` passthroughs in `marketplace_service.py` are removed, `create_mock_signer` is consolidated into `tests/unit/helpers.py` and reused across the three test files that had inline duplication, `test_preserves_explicit_zero_gas` covers the `gas=0` fix, `test_real_key_signature_recovers_via_raw_ecrecover` is exactly the crypto-level regression I was after (constructs a real `EthereumCrypto`, verifies raw-digest `ecrecover` recovery), and `test_auto_deposit_native` now asserts `mock_ds.assert_called_once_with(..., signer=service.signer, ...)` so a future refactor dropping the signer forwarding would fail loudly. Approving.

mech_client/services/marketplace_service.py
```diff
@@ -410,10 +413,9 @@ async def _send_offchain_request(  # pylint: disable=too-many-arguments,too-many
             request_id_int = int.from_bytes(request_id_bytes, byteorder="big")
             request_id_hex = request_id_bytes.hex()
 
-            # Sign the request ID
-            signature = self.crypto.sign_message(
-                request_id_bytes, is_deprecated_mode=True
-            )
+            # Sign the request ID digest (raw ecrecover on-chain, no EIP-191
+            # prefix — see Signer.sign_message)
+            signature = "0x" + self.signer.sign_message(request_id_bytes).hex()
```
jmoreira-valory: **LOW (follow-up idea) — defensive `ecrecover` at fire time to catch EIP-191 external-signer mismatch.**

The `Signer.sign_message` docstring is explicit and correct: raw-digest signing (`unsafe_sign_hash` semantics), no EIP-191 personal-message prefix, because the on-chain `MechMarketplace` verifier does plain `ecrecover(digest, v, r, s)`. `LocalSigner` gets this right (`is_deprecated_mode=True`).

The risk is for external `Signer` implementers (Pearl BYOA, key-in-signer-service setups — the whole point of this PR): the obvious default in third-party signer libs (`eth_account.Account.sign_message`) applies EIP-191 automatically, producing signatures that ecrecover to a *different* address than the signer's `.address`. The Marketplace contract's `ecrecover` still succeeds (it just recovers the wrong address) so the failure is silent on-chain — the request just gets rejected as "not from a paying sender", and only debug-tracing reveals the recovered address is off.

A cheap belt-and-braces guard would be to recover the signature locally and compare against `self.signer.address` immediately after signing, before the payload goes out:

```python
from eth_account.messages import _hash_eip191_message  # or eth_utils / eth_keys
from eth_keys import KeyAPI

signature = "0x" + self.signer.sign_message(request_id_bytes).hex()
# Verify the signer used raw-digest signing, not EIP-191
sig = KeyAPI().Signature(hexstr=signature)
recovered = sig.recover_public_key_from_msg_hash(request_id_bytes).to_checksum_address()
if recovered.lower() != self.signer.address.lower():
    raise ValueError(
        f"Signer produced a signature that recovers to {recovered}, "
        f"not the signer address {self.signer.address}. "
        f"Most likely the signer applied EIP-191; the marketplace "
        f"contract needs raw-digest signing (see Signer.sign_message)."
    )
```

This catches the misconfiguration at fire time with a clear diagnostic, rather than at on-chain settle time with an opaque "not paid" state. Non-blocking; the docstring warning may well be enough — worth deciding whether the fleet-wide UX cost of a silent failure justifies the ~5 lines of belt-and-braces here.

jmoreira-valory (dismissed):
Independent review at HEAD `a9e270c2`. Verified the pre-validated Safe signature encoding (`12 zero + 20 owner + 32 zero + 0x01` = 65 B, matches the Safe standard exactly, valid because outer sender is the encoded owner), byte-for-byte offchain compatibility (`"0x" + signer.sign_message(bytes).hex()` produces the same 132-char hex as the old `crypto.sign_message(bytes, is_deprecated_mode=True)`), and the `LocalSigner` defaults (nonce/gasPrice/gas fill). The `is_deprecated_mode=True` semantics correctly bypass EIP-191 to match the on-chain raw `ecrecover` verifier.

@DIvyaNautiyal07's P1/P2 round is well-resolved — the dead-state cleanup landed, `if not tx.get("gas")` → `if tx.get("gas") is None` fixed the `gas=0` falsy trap, and the author's source-level web3 7.15.0 verification of `TRANSACTION_DEFAULTS` on the nonce comment is correct (`build_transaction` cannot fill nonce because `nonce` is not in that dict).

One non-blocking follow-up idea inline. LGTM.

mech_client/services/marketplace_service.py
OjusWiZard: Added in 3f3213e — agreed the silent on-chain failure mode is worth the ~5 lines, given external implementers are the whole audience here.

Implementation notes vs. the sketch:
- Used `Account._recover_hash` instead of `eth_keys.KeyAPI` — `KeyAPI.Signature` requires `v ∈ {0, 1}`, so raw signer output with `v ∈ {27, 28}` (e.g. `LocalSigner`) would need inverse normalization anyway. The guard normalizes `v` for the local check only and sends the original bytes unchanged (the contract accepts both encodings).
- Also rejects wrong-length and unrecoverable (e.g. `r = s = 0`) signatures with distinct diagnostics, so every malformed-signer case fails at fire time with a message pointing at `Signer.sign_message`.
- The offchain tests now sign with a real key (`Account.unsafe_sign_hash`) so the guard is exercised end-to-end, plus dedicated tests for the EIP-191, low-`v`, wrong-length, and unrecoverable branches.

jmoreira-valory (approved):
Recheck at HEAD `3f3213e76b70eb8e884a5e070eea12c8e31061b7`. My prior LOW #2 (defensive `ecrecover` guard against EIP-191 external-signer mismatch) landed in `3f3213e` — cleaner than the sketch I suggested:

- Extracted into a dedicated `_sign_request_digest` helper on `MarketplaceService` (better separation of concerns than inlining it at the callsite).
- Length check first (`len(signature) != 65`) catches malformed signer output before the recovery step.
- `v` byte {0, 1} → {27, 28} normalization for **recovery only** — original bytes go on the wire, matching the contract that accepts both encodings. Nice touch.
- `Account._recover_hash` over `eth_keys.KeyAPI` — the author's rationale (KeyAPI.Signature needs more manual assembly, `_recover_hash` is private-but-stable) is sound.
- Diagnostic messages specifically call out EIP-191 as the most likely cause, so the error goes straight to the misconfiguration when it fires.
- Three focused tests: `test_valid_signature_passes_and_is_hex_encoded`, `test_low_v_signature_is_normalized_for_recovery_only`, `test_eip191_signature_is_rejected_with_diagnostic`.

Also re-verified everything I approved at `a9e270c2` remains intact (pre-validated Safe signature encoding, byte-for-byte offchain compat, LocalSigner nonce/gas/gasPrice fill semantics, dead-state cleanup from DivyaNautiyal07's P1 round).

CI 30/30 green. LGTM.

DIvyaNautiyal07 (approved):

PR #249, opened 2026-07-24
feat(marketplace): bind offchain requester to the Safe in agent mode

### LINKED ISSUE #248: Agent-mode auto_deposit credits the Safe's prepaid balance, but off-chain requests are charged to the signing EOA
**Summary**

In agent mode, the off-chain request and its auto-deposit top-up use two different requester identities, so the top-up can never clear the 402 it was made for:

- The off-chain request is authenticated by plain `ecrecover` over the request id, so the request `sender` is necessarily the **agent EOA** (`marketplace_service._send_offchain_request`: `sender = self.signer.address`). The mech checks `mapRequesterBalances[EOA]` on the balance tracker.
- `auto_deposit` / `DepositService` in agent mode executes through the **Safe** (`AgentExecutor` → `SafeClient.send_transaction`), and the tracker's `receive()` / `deposit()` credit `msg.sender` — i.e. `mapRequesterBalances[safe]`.

Even with a well-formed 402 challenge, the flow would be: deposit (credits the safe) → retry → mech still sees the EOA at 0 → second 402 → `"Auto-deposit did not clear the 402"` raised *after* funds already moved out of the safe.

**Verified on-chain** (Gnosis fork, `BalanceTrackerFixedPriceNative` `0x21cE6799A22A3Da84B7c44a814a9c79ab1d2A50D`, mech-client 0.21.2): an agent-mode `DepositService.deposit_native(10**16)` moved `mapRequesterBalances[safe]` 0 → 10¹⁶ while `mapRequesterBalances[EOA]` stayed 0, and the mech kept rejecting; a plain EOA→tracker transfer (crediting the EOA) let the same request be served and delivered.

**Possible fixes**

1. In the off-chain flow, make auto-deposit pay from the EOA (client-mode executor) so the credited account matches the charged one — with the caveat that agent EOAs typically hold only gas money;
2. a `depositFor(beneficiary)`-style tracker call so the safe can fund the EOA's balance; or
3. resolve EOA→safe on the mech/tracker side so the safe's balance backs the EOA's signed requests (arguably the right model for Olas services, where funds live in the service safe by design).

Until then, agent-mode integrators should treat off-chain prepayment as an operator action on the EOA's own balance and disable `auto_deposit`.

Related: the 402 body itself currently carries no amounts for auto-deposit to act on — valory-xyz/mech#469.

DIvyaNautiyal07:
## Summary

Aligns mech-client's offchain marketplace flow with the onchain flow. In agent mode, the Safe multisig is now the requester of record on every surface that keys on identity:

- Payload `sender` field posted to the mech server
- `mapNonces(sender)` lookup on the marketplace
- `getRequestId(mech, sender, ...)` computation
- On-chain signature verification (`_verifySignedHash` in MechMarketplace)

Client mode is unchanged and continues to bind the EOA.

## Why this change

Before this PR, the offchain flow used `sender = self.signer.address` (the EOA). Two problems:

1. **Internal inconsistency with the onchain flow.** In agent mode, onchain requests already have the Safe as msg.sender, so the marketplace records `requester = Safe`, and deposits (also from the Safe via the executor) credit `mapRequesterBalances[Safe]`. Offchain declared the EOA as sender, so the mech server pre-flighted `mapRequesterBalances[EOA]` which was always 0. Every agent-mode offchain request would hit HTTP 402 forever, because auto-deposit still credits the Safe slot (the executor sends from the Safe).
2. **Inconsistency with the Autonolas service model.** A service is represented by a Safe multisig. Every other component in the ecosystem (the trader via mech-interact, the marketplace, the subgraph's aggregation over on-chain traffic) treats the Safe as the entity that owns a request. Offchain being EOA-based was an outlier.

Live subgraph data confirms the direction. On Gnosis marketplace-gnosis subgraph, all 10 most recent `MarketplaceRequest` events (onchain path) have Safe requesters with threshold 1. The 41 `MarketplaceDeliveryWithSignatures` events that exist across all history (offchain-signed path) all have EOA requesters, and are entirely test volume from mech-client. Aligning mech-client's offchain path to Safe-as-sender fits both the production convention and the ecosystem model.

## What changed

- `Signer.sign_safe_message(safe_address, chain_id, message) -> bytes` added to the signer protocol. Downstream `Signer` implementers (e.g. Pearl BYOA) must add this method to remain compatible with agent-mode offchain requests. Client mode is served by the existing `sign_message` and requires no changes.
- `LocalSigner.sign_safe_message` computes the Safe v1.4.1 SafeMessage EIP-712 wrapped hash locally (no RPC round-trip to `Safe.getMessageHash`) and signs it raw. Documented Safe v1.4.1 assumption (chainId in the EIP-712 domain).
- `MarketplaceService._resolve_offchain_sender()` returns the Safe in agent mode and the EOA in client mode. All identity-keyed surfaces (nonce, request-id, payload sender, signature) bind to the same address per mode.
- `MarketplaceService._sign_request_digest` is mode-conditional: agent mode routes through `sign_safe_message`, client mode keeps the raw-digest path.
- Deposit flow left alone (already credits `msg.sender = Safe` in agent mode, which is now the requester of record).
- Version bumps: mech-client `0.21.2` -> `0.21.3`, `open-autonomy 0.21.25` -> `0.21.26`, `open-aea-helpers 0.21.25` -> `0.21.26`.

## Signature verification, and why two formats

`MechMarketplace._verifySignedHash` branches on the requester's bytecode:

```solidity
if (requester.code.length > 0) {
    // Safe branch: EIP-1271 magic value check
    ISignatureValidator(requester).isValidSignature(request_id, sig) == MAGIC_VALUE
} else {
    // EOA branch: plain ecrecover on the raw digest
    require(ecrecover(request_id, v, r, s) == requester);
}
```

There is no single signature format that both branches accept. Safe's `CompatibilityFallbackHandler` (v1.4.1) wraps the raw `request_id` into a `SafeMessage` EIP-712 struct bound to `(chainId, safeAddress)` and validates against the wrapped hash. A raw-ECDSA signature over the unwrapped digest reverts with `GS026`. So the client signs the raw digest in client mode and the wrapped hash in agent mode. Same signing key, same signing function underneath. Only the 32-byte input differs.

## Test plan

- [x] Unit tests for `LocalSigner.sign_safe_message`: fork-verified reference vector, invariants across Safe/chainId/message inputs.
- [x] Unit tests for `MarketplaceService._resolve_offchain_sender`: agent vs client mode, missing-Safe error.
- [x] Unit tests for `MarketplaceService._sign_request_digest`: agent-mode wrapper, length guard.
- [x] Integration-style tests: end-to-end offchain request binds EOA in client mode and Safe in agent mode (mapNonces, getRequestId, payload sender all consistent).
- [x] Fork verification (Gnosis mainnet via anvil): freshly-deployed Safe v1.4.1 returns `0x1626ba7e` from `isValidSignature` for signatures produced by `LocalSigner.sign_safe_message`.
- [x] Full CI locally: black, isort, flake8, mypy, pylint (10.00/10), darglint, bandit, liccheck, check-third-party-hashes, check-copyright, gitleaks, uv lock check.
- [x] 767 tests pass, 100% coverage maintained.

## Follow-up

`valory-xyz/mech-interact#112` applies the same SafeMessage-wrapped signing to `_sign_request_id` so their existing Safe-as-sender offchain path can settle on-chain.

Fixes #248.

### DIFF
--- a/mech_client/domain/signing/base.py
+++ b/mech_client/domain/signing/base.py
@@ -75,7 +75,7 @@ class Signer(Protocol):
         Used by the agent-mode offchain marketplace flow, where the requester
         of record is a Safe multisig (not the signing EOA). The marketplace
         verifies via ``Safe.isValidSignature(digest, sig)``, which on Safe
-        v1.4.1 with the standard ``CompatibilityFallbackHandler`` rehashes
+        v1.3.0+ with the standard ``CompatibilityFallbackHandler`` rehashes
         the raw ``digest`` into an EIP-712 ``SafeMessage`` and checks that
         the sig recovers to an owner over the wrapped hash. A raw
         ``ecrecover`` signature over ``digest`` fails on-chain with
@@ -91,17 +91,32 @@ class Signer(Protocol):
         to ``Safe.getMessageHash``) and sign it raw (``v`` in ``{27, 28}``).
         The signer must be an owner of ``safe_address``.
 
+        ``message`` MUST be exactly 32 bytes. Callers pass the request-id
+        digest; the wrapping formula's shortcut ``keccak256(message)`` only
+        equals ``keccak256(abi.encode(bytes32, message))`` at 32 bytes, so
+        any other length silently signs a hash the fallback handler will
+        not accept. Implementations MUST reject non-32-byte inputs.
+
         This is a required protocol member. Downstream implementers of
         :class:`Signer` (e.g. Pearl BYOA agents) must add this method to
         remain compatible with agent-mode offchain requests. Existing
         :meth:`sign_message` is unchanged and continues to serve client mode.
 
-        Domain assumption: Safe v1.4.1. Older Safes (v1.3.0) use a domain
-        typehash without ``chainId`` and are not supported by this method.
+        Domain assumption: Safe v1.3.0+ (v1.3.0 and v1.4.1 share the
+        ``EIP712Domain(uint256 chainId,address verifyingContract)``
+        typehash and the same ``SafeMessage`` wrapping). ``mechx setup``
+        deploys the canonical Safe v1.3.0 singleton, so this method
+        targets that domain. Pre-v1.3.0 Safes (v1.2.0 and earlier) use a
+        domain typehash without ``chainId`` and are not supported.
+
+        The Safe's signature threshold must be 1 for this single-owner
+        signature to validate. Safes with threshold > 1 will reject the
+        signature via ``checkNSignatures`` (a distinct failure mode from
+        ``GS026``).
 
         :param safe_address: Checksummed Safe address (verifyingContract in
             the EIP-712 domain)
         :param chain_id: Chain ID of the network the Safe is deployed on
-        :param message: Message bytes to wrap and sign (32-byte digest for
-            mech requests)
+        :param message: 32-byte digest to wrap and sign (the mech
+            request-id digest)
         """
--- a/mech_client/domain/signing/base.py
+++ b/mech_client/domain/signing/base.py
@@ -57,12 +57,51 @@ class Signer(Protocol):
     def sign_message(self, message: bytes) -> bytes:
         """Return a 65-byte ECDSA signature (r ‖ s ‖ v) over ``message``.
 
-        The mech marketplace flow passes the raw 32-byte request-id digest
-        here, and the on-chain verifier recovers it with plain
-        ``ecrecover(digest, v, r, s)`` — i.e. the digest must be signed
-        directly (``eth_account``'s ``unsafe_sign_hash`` semantics), NOT
-        wrapped in an EIP-191 personal-message prefix. ``v`` may be
+        Used by the client-mode offchain marketplace flow: the EOA is the
+        requester of record, and the marketplace verifier recovers with
+        plain ``ecrecover(digest, v, r, s)`` — i.e. the digest must be
+        signed directly (``eth_account``'s ``unsafe_sign_hash`` semantics),
+        NOT wrapped in an EIP-191 personal-message prefix. ``v`` may be
         ``{0, 1}`` or ``{27, 28}``; the contract normalizes both.
 
         :param message: Message bytes to sign (32-byte digest for mech requests)
+        """
+
+    def sign_safe_message(
+        self, safe_address: str, chain_id: int, message: bytes
+    ) -> bytes:
+        """Return a 65-byte ECDSA signature over the Safe-wrapped ``message``.
+
+        Used by the agent-mode offchain marketplace flow, where the requester
+        of record is a Safe multisig (not the signing EOA). The marketplace
+        verifies via ``Safe.isValidSignature(digest, sig)``, which on Safe
+        v1.4.1 with the standard ``CompatibilityFallbackHandler`` rehashes
+        the raw ``digest`` into an EIP-712 ``SafeMessage`` and checks that
+        the sig recovers to an owner over the wrapped hash. A raw
+        ``ecrecover`` signature over ``digest`` fails on-chain with
+        ``GS026``; the signature MUST be produced over::
+
+            wrappedHash = keccak256(
+                0x1901
+                || keccak256(abi.encode(DOMAIN_TYPEHASH, chainId, safe))
+                || keccak256(abi.encode(SAFE_MSG_TYPEHASH, keccak256(abi.encode(message))))
+            )
+
+        Implementations MUST compute ``wrappedHash`` locally (no RPC call
+        to ``Safe.getMessageHash``) and sign it raw (``v`` in ``{27, 28}``).
+        The signer must be an owner of ``safe_address``.
+
+        This is a required protocol member. Downstream implementers of
+        :class:`Signer` (e.g. Pearl BYOA agents) must add this method to
+        remain compatible with agent-mode offchain requests. Existing
+        :meth:`sign_message` is unchanged and continues to serve client mode.
+
+        Domain assumption: Safe v1.4.1. Older Safes (v1.3.0) use a domain
+        typehash without ``chainId`` and are not supported by this method.
+
+        :param safe_address: Checksummed Safe address (verifyingContract in
+            the EIP-712 domain)
+        :param chain_id: Chain ID of the network the Safe is deployed on
+        :param message: Message bytes to wrap and sign (32-byte digest for
+            mech requests)
         """
--- a/mech_client/domain/signing/local.py
+++ b/mech_client/domain/signing/local.py
@@ -22,6 +22,18 @@
 from typing import Any, Dict
 
 from aea_ledger_ethereum import EthereumApi, EthereumCrypto
+from eth_abi import encode as abi_encode
+from eth_account import Account
+from eth_utils import keccak
+from mech_client.utils.validators import ensure_checksummed_address
+
+# EIP-712 typehash constants for Safe v1.4.1 SafeMessage wrapping.
+# Recomputed as keccak of the type string below so a reader can diff them
+# against the on-chain constants without decoding hex.
+_EIP712_DOMAIN_TYPEHASH = keccak(
+    text="EIP712Domain(uint256 chainId,address verifyingContract)"
+)
+_SAFE_MESSAGE_TYPEHASH = keccak(text="SafeMessage(bytes message)")
 
 
 class LocalSigner:
@@ -87,3 +99,45 @@ class LocalSigner:
         """
         signature = self.crypto.sign_message(message, is_deprecated_mode=True)
         return bytes.fromhex(signature.removeprefix("0x"))
+
+    def sign_safe_message(
+        self, safe_address: str, chain_id: int, message: bytes
+    ) -> bytes:
+        """Sign the Safe-wrapped EIP-712 hash of ``message`` with the local key.
+
+        Computes the SafeMessage wrapped hash locally (no RPC) and signs it
+        raw so ``Safe.isValidSignature(message, sig)`` returns the ERC-1271
+        magic value on-chain — see :meth:`Signer.sign_safe_message` for the
+        formula and domain assumptions.
+
+        :param safe_address: Safe (verifyingContract) address
+        :param chain_id: Chain ID for the EIP-712 domain
+        :param message: Message bytes to wrap and sign
+        :return: 65-byte signature (r ‖ s ‖ v) over the wrapped hash
+        """
+        checksummed_safe = ensure_checksummed_address(safe_address)
+        domain_separator = keccak(
+            abi_encode(
+                ["bytes32", "uint256", "address"],
+                [_EIP712_DOMAIN_TYPEHASH, chain_id, checksummed_safe],
+            )
+        )
+        # ``Safe.isValidSignature(bytes32 hash, bytes sig)`` on the
+        # v1.4.1 CompatibilityFallbackHandler wraps ``hash`` as
+        # ``keccak256(abi.encode(hash))`` before hashing into
+        # ``SafeMessage``. For a 32-byte input, ``abi.encode(bytes32)``
+        # yields the same 32 bytes, so ``keccak256(message)`` here is
+        # exactly ``keccak256(<32 raw bytes>)`` — the same value the
+        # handler computes on-chain.
+        struct_hash = keccak(
+            abi_encode(
+                ["bytes32", "bytes32"],
+                [_SAFE_MESSAGE_TYPEHASH, keccak(message)],
+            )
+        )
+        wrapped_hash = keccak(b"\x19\x01" + domain_separator + struct_hash)
+        # pylint: disable-next=no-value-for-parameter
+        signed = Account.unsafe_sign_hash(
+            wrapped_hash, private_key=self.crypto.private_key
+        )
+        return bytes(signed.signature)
--- a/mech_client/domain/signing/local.py
+++ b/mech_client/domain/signing/local.py
@@ -27,13 +27,20 @@ from eth_account import Account
 from eth_utils import keccak
 from mech_client.utils.validators import ensure_checksummed_address
 
-# EIP-712 typehash constants for Safe v1.4.1 SafeMessage wrapping.
-# Recomputed as keccak of the type string below so a reader can diff them
-# against the on-chain constants without decoding hex.
+# EIP-712 typehash constants for Safe v1.3.0+ SafeMessage wrapping (the
+# chainId-in-domain typehash was introduced in v1.3.0 and unchanged in
+# v1.4.1). Recomputed as keccak of the type string below so a reader can
+# diff them against the on-chain constants without decoding hex.
 _EIP712_DOMAIN_TYPEHASH = keccak(
     text="EIP712Domain(uint256 chainId,address verifyingContract)"
 )
 _SAFE_MESSAGE_TYPEHASH = keccak(text="SafeMessage(bytes message)")
+
+# ``sign_safe_message`` shortcut ``keccak256(message)`` for the inner
+# ``keccak256(abi.encode(bytes32, message))`` only holds at exactly 32
+# bytes. Any other length silently signs a hash Safe.isValidSignature
+# will reject.
+_SAFE_MESSAGE_REQUIRED_LEN = 32
 
 
 class LocalSigner:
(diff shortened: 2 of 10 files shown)


mech_client/domain/signing/base.py
```diff
@@ -57,12 +57,51 @@ def send_transaction(self, unsigned_tx: Dict[str, Any]) -> str:
...
+        Implementations MUST compute ``wrappedHash`` locally (no RPC call
+        to ``Safe.getMessageHash``) and sign it raw (``v`` in ``{27, 28}``).
+        The signer must be an owner of ``safe_address``.
+
+        This is a required protocol member. Downstream implementers of
+        :class:`Signer` (e.g. Pearl BYOA agents) must add this method to
+        remain compatible with agent-mode offchain requests. Existing
+        :meth:`sign_message` is unchanged and continues to serve client mode.
+
+        Domain assumption: Safe v1.4.1. Older Safes (v1.3.0) use a domain
```
OjusWiZard: **[Important] The version claim is backwards — and v1.3.0 is what `mechx setup` actually deploys.**

Two issues with "Older Safes (v1.3.0) use a domain typehash without `chainId` and are not supported":

1. `chainId` was **added** to the Safe EIP-712 domain typehash in v1.3.0 (as the cross-chain-replay fix); it's v1.2.0 and earlier that lack it. v1.3.0 and v1.4.1 share the exact domain typehash implemented here, and v1.3.0's `CompatibilityFallbackHandler` uses the same `SafeMessage(bytes message)` wrapping — so signatures very likely already work on v1.3.0.
2. The practical stakes: `mechx setup` → `olas-operate-middleware` → `packages/valory/contracts/gnosis_safe/contract.py` deploys the canonical **v1.3.0** singleton (`0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552`, listed as `"1.3.0"` in `safe_eth.safe.addresses.MASTER_COPIES`) on every supported chain. So the fork vector — the PR's only byte-level proof — was captured against a *different* Safe version than every real agent-mode user will have.

Suggested fix: correct this docstring (supported = v1.3.0+, unsupported = v1.2.0 and earlier) and regenerate/add a fork vector against the v1.3.0 deployment `mechx setup` produces.

mech_client/domain/signing/base.py
```diff
@@ -57,12 +57,51 @@ def send_transaction(self, unsigned_tx: Dict[str, Any]) -> str:
...
+                0x1901
+                || keccak256(abi.encode(DOMAIN_TYPEHASH, chainId, safe))
+                || keccak256(abi.encode(SAFE_MSG_TYPEHASH, keccak256(abi.encode(message))))
+            )
+
+        Implementations MUST compute ``wrappedHash`` locally (no RPC call
+        to ``Safe.getMessageHash``) and sign it raw (``v`` in ``{27, 28}``).
+        The signer must be an owner of ``safe_address``.
+
+        This is a required protocol member. Downstream implementers of
```
OjusWiZard, line 100: **[Important] Pre-PR downstream `Signer` implementations fail with a bare `AttributeError` in agent mode.**

`_sign_request_digest` calls `self.signer.sign_safe_message(...)` unconditionally when `agent_mode` is true, so a Pearl BYOA signer built against the old protocol dies with `AttributeError` deep inside the request flow — inconsistent with the repo convention of actionable error messages (CLAUDE.md §2). The break itself is documented here, so this is about error quality, not the decision: a `hasattr` guard at the call site raising something like *"your Signer must implement sign_safe_message for agent-mode offchain requests — see Signer.sign_safe_message"* costs three lines.

mech_client/domain/signing/base.py
```diff
@@ -57,12 +57,51 @@ def send_transaction(self, unsigned_tx: Dict[str, Any]) -> str:
...
+        requester of record, and the marketplace verifier recovers with
+        plain ``ecrecover(digest, v, r, s)`` — i.e. the digest must be
+        signed directly (``eth_account``'s ``unsafe_sign_hash`` semantics),
+        NOT wrapped in an EIP-191 personal-message prefix. ``v`` may be
         ``{0, 1}`` or ``{27, 28}``; the contract normalizes both.
 
         :param message: Message bytes to sign (32-byte digest for mech requests)
         """
+
+    def sign_safe_message(
```
OjusWiZard, line 70: **[Advisory] Undocumented threshold-1 assumption.**

The flow sends exactly one owner signature as the entire payload; a Safe with threshold > 1 will reject it via `checkNSignatures` — a different failure mode from the documented `GS026` case. Since the PR body itself notes production requesters are threshold-1, one sentence here stating the precondition ("the Safe's signature threshold must be 1") would save the next debugging session.

mech_client/domain/signing/local.py
```diff
@@ -87,3 +99,45 @@ def sign_message(self, message: bytes) -> bytes:
         """
         signature = self.crypto.sign_message(message, is_deprecated_mode=True)
         return bytes.fromhex(signature.removeprefix("0x"))
+
+    def sign_safe_message(
```
OjusWiZard, line 110: **[Important] No 32-byte guard on `message` — any other length silently signs a wrong hash.**

The comment below correctly notes the `keccak(message)` shortcut is only equivalent to on-chain `keccak256(abi.encode(bytes32))` for exactly 32-byte inputs, but nothing enforces that. Verified empirically: an empty or 31-byte message is accepted and yields a well-formed 65-byte signature that can only fail later as an opaque on-chain `GS026` — the exact silent failure mode the sibling `sign_message` path goes out of its way to prevent.

```python
if len(message) != 32:
    raise ValueError(
        f"sign_safe_message requires a 32-byte digest, got {len(message)} bytes"
    )
```

Worth also stating the 32-byte precondition as a hard requirement in the `Signer.sign_safe_message` protocol docstring so third-party implementers enforce it too.

OjusWiZard (commented):
Multi-perspective review (correctness, comments, tests, silent failures, type design, simplification). **No Critical findings; 4 Important, 5 Advisory — posted inline.**

What reviewed clean:
- The EIP-712 wrapped-hash construction was independently re-derived with `eth_account.messages.encode_typed_data` (a separate encoder from the manual `abi_encode`/`keccak` path) and matched the fork-verified reference signature byte-for-byte. Typehash constants match Safe's on-chain values.
- Mode-conditional sender binding is correct and well-tested on all three identity surfaces (`mapNonces`, `getRequestId`, payload `sender`); the tests deliberately use distinct EOA/Safe addresses so a regression would surface. Client mode is genuinely unchanged.
- 767 tests pass / 100% coverage reproduced locally; darglint-relevant docstring sections match signatures; no unsafe key handling.
- "Agent mode without Safe" is already unrepresentable via the public constructor (`ExecutorFactory.create` raises), so the new guard in `_resolve_offchain_sender` is sound defense-in-depth.

Suggested priority: the Safe-version finding first (only one with likely real-user impact today), then the two fail-fast guards, then error quality for downstream signers. Advisories are optional polish.

mech_client/domain/signing/local.py
```diff
@@ -22,6 +22,18 @@
...
 from aea_ledger_ethereum import EthereumApi, EthereumCrypto
+from eth_abi import encode as abi_encode
+from eth_account import Account
+from eth_utils import keccak
+from mech_client.utils.validators import ensure_checksummed_address
+
+# EIP-712 typehash constants for Safe v1.4.1 SafeMessage wrapping.
+# Recomputed as keccak of the type string below so a reader can diff them
+# against the on-chain constants without decoding hex.
+_EIP712_DOMAIN_TYPEHASH = keccak(
```
OjusWiZard, line 34: **[Advisory] The manual EIP-712 plumbing could be `eth_account.messages.encode_typed_data`.**

Verified byte-identical against this PR's own fork vector (including with a lowercase address input):

```python
signable = encode_typed_data(
    {"chainId": chain_id, "verifyingContract": checksummed_safe},
    {"SafeMessage": [{"name": "message", "type": "bytes"}]},
    {"message": message},
)
signed = Account.sign_message(signable, private_key=self.crypto.private_key)
```

Still purely local (no RPC), removes ~15 lines of hand-rolled ABI/keccak plumbing and both typehash constants. Legitimate counter-argument: the current verbose form is a deliberate audit trail matching the formula in `base.py` — author's call, flagging as an option rather than a request.

mech_client/services/marketplace_service.py
```diff
@@ -466,24 +472,65 @@ async def _send_offchain_request(  # pylint: disable=too-many-arguments,too-many
...
+
         :param request_id_bytes: the 32-byte request-id digest to sign.
         :return: 0x-prefixed hex signature, as sent to the mech.
-        :raises ValueError: if the signature has the wrong length or does not
-            recover to the signer's address (e.g. the signer applied EIP-191).
+        :raises ValueError: if the signature has the wrong length or
+            (client mode) does not recover to the signer's address.
         """
+        if self.agent_mode:
+            signature = self.signer.sign_safe_message(
```
OjusWiZard, line 538: **[Important] Agent-mode branch validates only length — a wrong `v` byte is misinterpreted, not rejected.**

The client-mode branch below recovers the signature locally and compares against `signer.address`, converting signer misconfiguration into an immediate error. This branch accepts any 65 bytes. A third-party `Signer` returning `v ∈ {0, 1}` (a common ECDSA convention, and explicitly tolerated by `sign_message`'s contract) is dangerous here: Safe's `checkNSignatures` interprets `v=0` as a contract signature and `v=1` as an approved-hash, so the signature wouldn't just fail — it would be misinterpreted, surfacing only as `GS026`-style rejection on-chain.

Full Safe-ownership verification would need an RPC call (documented design tradeoff, fine to skip), but the cheap check closes the worst case:

```python
if signature[64] not in (27, 28):
    raise ValueError(
        f"Signer returned v={signature[64]}; sign_safe_message requires "
        f"raw-hash signing with v in {{27, 28}} — see Signer.sign_safe_message."
    )
```

mech_client/services/marketplace_service.py
```diff
@@ -466,24 +472,65 @@ async def _send_offchain_request(  # pylint: disable=too-many-arguments,too-many
...
+                self._resolve_offchain_sender(),
```
OjusWiZard: **[Advisory] `_resolve_offchain_sender()` is re-derived here on every prompt.**

`_send_offchain_request` already resolved and holds `sender` before the loop. Passing it in as a parameter (`_sign_request_digest(self, sender: str, request_id_bytes: bytes)`) makes the invariant "the address the signature is computed against == the address in the payload/`getRequestId`" structural instead of two call sites happening to compute the same value. Two review passes flagged this independently. Mechanical test-callsite updates, otherwise zero cost.

tests/unit/services/test_marketplace_service.py
```diff
@@ -1869,6 +2132,34 @@ def _build_offchain_service() -> MarketplaceService:
         )
 
 
+def _build_offchain_service_agent_mode(
```
OjusWiZard: **[Advisory]** This helper duplicates `_build_offchain_service` verbatim except for the constructor kwargs — same 5-patch `with` block. They collapse into one helper, e.g. `_build_offchain_service(agent_mode: bool = False, safe_address: Optional[str] = None)`, that conditionally adds `safe_address`/`ethereum_client`. Test-only, no impact on the coverage/lint gates.

tests/unit/domain/test_signing.py
```diff
@@ -231,3 +237,163 @@ def test_real_key_signature_recovers_via_raw_ecrecover(
...
+
+        signature = signer.sign_safe_message(
+            _SAFE_FORK_SAFE_ADDRESS,
+            _SAFE_FORK_CHAIN_ID,
+            _SAFE_FORK_REQUEST_ID,
+        )
+
+        assert isinstance(signature, bytes)
+
+    def test_different_safe_address_yields_different_signature(
```
OjusWiZard, line 357: **[Advisory]** This test and the two below (`chain_id`, `message`) share an identical shape: sign the reference vector, sign a one-field variant, assert the signatures differ. They collapse into a single `@pytest.mark.parametrize` with `ids=["safe_address", "chain_id", "message"]` (~45 → ~15 lines) while keeping per-case failure attribution. The per-case replay-rationale docstrings can be folded into the combined docstring. Minor — the tests themselves are genuinely good differential tests.

mech_client/domain/signing/base.py
DIvyaNautiyal07: Corrected in e60aaf2. Docstrings across base.py, local.py, marketplace_service.py, and the test module now say v1.3.0+ (the chainId-in-domain typehash was introduced in v1.3.0 and is unchanged in v1.4.1). Re-verified on an anvil Gnosis fork against the canonical v1.3.0 singleton `0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552` that `mechx setup` deploys: `isValidSignature` returns `0x1626ba7e`. The v1.3.0 signature is pinned as a second reference vector alongside the v1.4.1 one, so a future regression breaking either version fails the suite.

mech_client/domain/signing/base.py
DIvyaNautiyal07: Fixed in e60aaf2. `_sign_request_digest` now checks `hasattr(self.signer, "sign_safe_message")` in the agent-mode branch and raises `ValueError` with a message referencing `Signer.sign_safe_message` before the flow reaches the call. Downstream implementers get a clear error at fire time instead of an `AttributeError` deep in the request path.

mech_client/domain/signing/base.py
DIvyaNautiyal07: Added in e60aaf2. `Signer.sign_safe_message` docstring now states the threshold-1 precondition and calls out that Safes with threshold > 1 will reject the single-owner signature via `checkNSignatures` (different failure mode from GS026).

mech_client/domain/signing/local.py
DIvyaNautiyal07: Fixed in e60aaf2. `LocalSigner.sign_safe_message` raises `ValueError` on `len(message) != 32`. The 32-byte precondition is also stated as a hard requirement in the `Signer.sign_safe_message` protocol docstring so third-party implementers enforce it too.

mech_client/domain/signing/local.py
DIvyaNautiyal07: Skipping per your own opt-out note. The hand-rolled ABI/keccak form is a deliberate audit trail that matches the wrapping formula in base.py line-for-line; the switch would be cosmetic.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Fixed in e60aaf2. Agent-mode branch now raises `ValueError` if `signature[64] not in (27, 28)` with a message referencing `Signer.sign_safe_message`. A signer returning v in {0, 1} now surfaces at the client instead of being misinterpreted by Safe as a contract-sig or approvedHash and failing as GS026 on-chain.

mech_client/services/marketplace_service.py
DIvyaNautiyal07: Fixed in e60aaf2. `_sign_request_digest` now takes `(sender, request_id_bytes)`; `_send_offchain_request` resolves `sender` once outside the loop and threads it in. New test `test_agent_mode_signs_against_caller_provided_sender` asserts the sign path uses the caller-provided value rather than re-reading `self.safe_address`, so the invariant is structural.

tests/unit/services/test_marketplace_service.py
DIvyaNautiyal07: Fixed in e60aaf2. `_build_offchain_service_agent_mode` collapsed into `_build_offchain_service(agent_mode=False, safe_address=None)`; the four call sites are updated.

tests/unit/domain/test_signing.py
DIvyaNautiyal07: Skipping. The three replay-rationale tests are already well-named and their explicit form makes each case's intent immediately readable; the collapse would be a per-line count improvement with no functional gain.

OjusWiZard (approved):
Approving after a two-round multi-perspective review.

Round 1 findings (Safe version accuracy, 32-byte digest guard, v-byte guard, downstream-signer error, threshold-1 docs, sender threading, test consolidation) were all addressed in e60aaf2 and verified against the diff — threads resolved.

Round 2 re-review of the hardening commit found no correctness or silent-failure issues: all guards fail loudly through to the CLI boundary, the v1.3.0 fork vector is genuine, and 780 tests pass at 100% coverage. Remaining minor items, none blocking:
- flake8 E302 at tests/unit/domain/test_signing.py:280 (one missing blank line — will fail the lint gate, worth fixing before merge)
- the v-byte guard's error text says v=0/v=1 "both fail as GS026"; actually GS022/GS025 respectively — GS026 is the raw-ecrecover branch
- no v=28 happy-path test at the MarketplaceService layer (fork vectors cover it only at the LocalSigner layer)
- `:raises` docstring on _sign_request_digest omits the missing-sign_safe_message case

kupermind (approved):

PR #251, opened 2026-08-11
fix(delivery): return the delivered content, in one shape (#250)

### LINKED ISSUE #250: delivery_results is unusable as returned: directory URL on-chain, parsed JSON off-chain
`delivery_results` can't be consumed as returned: the on-chain flow yields a directory URL, the off-chain flow yields parsed JSON.

Observed on `mech-client` 0.21.3, Gnosis.

## 1. The on-chain URL points at a directory, and the entry name is non-obvious

`OnchainDeliveryWatcher.watch_for_data_urls` builds the result as a bare gateway URL:

```python
# mech_client/domain/delivery/onchain_watcher.py:279
results[request_id] = IPFS_URL_TEMPLATE.format(delivery_data)
```

That CID is a directory, not the result file. Fetching what the caller is handed returns an HTML directory listing. The result file inside is named by the request id **in decimal** — the hex form that the `Deliver` event, the logs, and `request_ids` all use gives a 404.

Checked against three live deliveries on Gnosis (blocks 47633819, 47633845, 47633858; mech `0x601024E27f1C67B28209E24272CED8A31fc8151F`), consistent across all three:

| Fetch | Result |
|---|---|
| the URL as returned | `200 text/html` — "A directory of content-addressed files hosted on IPFS" |
| `<url>/0x15467d98…7582` (hex request id) | `404` |
| `<url>/9623115955…666370` (decimal request id) | `200 application/json` — the result |

So every caller has to independently rediscover the hex→decimal detail, or scrape the HTML listing to find the entry. `watch_for_data_urls` already has `request_id` in hand on that exact line, so it can build the full path itself.

There's a second unwrap after that. The payload is:

```json
{
  "schema_version": "2.0",
  "requestId": 9623115955867796542954572853826696724577646398124017908538111513641050666370,
  "result": "{\"p_yes\": 0.38, \"p_no\": 0.62, \"confidence\": 0.55, \"info_utility\": 0.9}",
  "tool": "factual_research",
  "executed_at": "2026-08-09T08:39:17.570740Z",
  "prompt": "..."
}
```

`result` is a JSON-encoded *string*, so getting to the answer needs a second `json.loads`.

## 2. The two flows return different types under the same key

- On-chain (`onchain_watcher.py:279`): `delivery_results[request_id]` is a URL `str`.
- Off-chain (`offchain_watcher.py:91`): `delivery_results[request_id]` is the parsed JSON response.

Both are returned under `delivery_results` from `MarketplaceService` (lines 343 and 471), so a caller that doesn't know which flow ran can't write one handler. Downstream code can paper over this, but only mech-client can actually fix it.

## Suggested fix

Have the on-chain watcher fetch the file it already knows the path to, and return the same shape the off-chain watcher does — parsed content, keyed by request id. If the raw URL is worth keeping, an adjacent key (`delivery_urls`, or a `{"url": ..., "data": ...}` value on both paths) would preserve it without the type split.

Happy to open a PR if this direction looks right.

OjusWiZard:
Fixes #250.

## The problem

`delivery_results[request_id]` had a different meaning depending on how the response arrived:

| Path | What you got |
|---|---|
| On-chain | An IPFS URL **string** |
| Off-chain | The mech's raw envelope **dict** |

So a caller could not write one handler for both.

The on-chain URL was also not usable as-is. A mech delivers by pinning a *directory*, so the hash in the `Deliver` event addresses the directory, not the answer — fetching it returns an HTML listing. The result file inside is named after the request ID **in decimal**, while the hex form used everywhere else in this client 404s.

## The change

Both watchers now build the full result-file path, read it, and return a `DeliveryResult` carrying the parsed content plus the URL it came from. `send_request` splits that into two keys with the same shape on both paths:

- `delivery_results[request_id]` — the parsed result file
- `delivery_urls[request_id]` — the gateway URL it was read from (`None` for off-chain mechs answering inline)

`mechx request` prints the mech's answer, decoding the JSON-encoded `result` field when the payload has one, followed by the result-file URL.

Result files are read **concurrently and off the event loop**. A batch delivers one file per request, and reading them serially would stack a gateway timeout per request *after* the caller's `--timeout` budget is already spent.

Unreadable files degrade rather than fail: `delivery_results` is `None` and the URL is still reported, so the answer stays retrievable by hand.

## Breaking change

Callers reading `delivery_results` as a URL must move to `delivery_urls`. Documented in the CHANGELOG, README, and `docs/index.md`.

## Verification

- 790 unit tests pass
- 100% line coverage on all touched modules; `request_cmd.py` at 100% **branch** coverage
- `black`, `isort`, `flake8`, `mypy`, `pylint`, `darglint`, `bandit`, and the copyright check all pass

### DIFF
--- a/mech_client/cli/commands/request_cmd.py
+++ b/mech_client/cli/commands/request_cmd.py
@@ -252,13 +252,11 @@ def request(
     # Display results
     click.echo(f"\n✓ Transaction hash: {result['tx_hash']}")
     click.echo(f"✓ Request IDs: {result['request_ids']}")
-    if result.get("delivery_results"):
-        delivery_urls = result.get("delivery_urls") or {}
+    if result.get("deliveries"):
         click.echo("\n✓ Delivery results:")
-        for request_id, delivery_data in result["delivery_results"].items():
+        for request_id, delivery in result["deliveries"].items():
             click.echo(
-                f"  Request {request_id}: {_format_delivery_output(delivery_data)}"
+                f"  Request {request_id}: {_format_delivery_output(delivery.data)}"
             )
-            url = delivery_urls.get(request_id)
-            if url:
-                click.echo(f"    Result file: {url}")
+            if delivery.url:
+                click.echo(f"    Result file: {delivery.url}")
--- a/mech_client/cli/commands/request_cmd.py
+++ b/mech_client/cli/commands/request_cmd.py
@@ -27,7 +27,6 @@ import click
 from click import ClickException
 from mech_client.cli.common import common_wallet_options, setup_wallet_command
 from mech_client.cli.validators import validate_chain_config, validate_ethereum_address
-from mech_client.infrastructure.config import IPFS_URL_TEMPLATE
 from mech_client.services.marketplace_service import MarketplaceService
 from mech_client.utils.errors.handlers import handle_cli_errors
 from mech_client.utils.validators import (
@@ -37,13 +36,40 @@ from mech_client.utils.validators import (
 )
 
 
+def _dump(value: Any) -> str:
+    """
+    Render a value parsed out of a delivered result file for CLI output.
+
+    :param value: A value decoded from JSON, so always JSON-serialisable.
+    :return: Indented JSON.
+    """
+    return json.dumps(value, ensure_ascii=True, indent=2, sort_keys=True)
+
+
 def _format_delivery_output(delivery_data: Any) -> str:
-    """Format delivery data for CLI output with parity across delivery modes."""
-    if isinstance(delivery_data, dict):
-        task_result = delivery_data.get("task_result")
-        if isinstance(task_result, str) and task_result:
-            return IPFS_URL_TEMPLATE.format(task_result)
-        return json.dumps(delivery_data, ensure_ascii=True, indent=2, sort_keys=True)
+    """
+    Format delivery data for CLI output with parity across delivery modes.
+
+    :param delivery_data: Parsed content of the delivered result file.
+    :return: The mech's answer, ready to print.
+    """
+    if delivery_data is None:
+        return "unavailable — could not read the result file (URL below)"
+
+    if isinstance(delivery_data, dict) and "result" in delivery_data:
+        result = delivery_data["result"]
+        # Mechs store `result` as a JSON-encoded string; decode it so the
+        # answer prints as itself rather than as an escaped blob.
+        if isinstance(result, str):
+            try:
+                result = json.loads(result)
+            except ValueError:
+                return result
+        return result if isinstance(result, str) else _dump(result)
+
+    if isinstance(delivery_data, (dict, list)):
+        return _dump(delivery_data)
+
     return str(delivery_data)
 
 
@@ -227,8 +253,12 @@ def request(
     click.echo(f"\n✓ Transaction hash: {result['tx_hash']}")
     click.echo(f"✓ Request IDs: {result['request_ids']}")
     if result.get("delivery_results"):
+        delivery_urls = result.get("delivery_urls") or {}
         click.echo("\n✓ Delivery results:")
         for request_id, delivery_data in result["delivery_results"].items():
             click.echo(
                 f"  Request {request_id}: {_format_delivery_output(delivery_data)}"
             )
+            url = delivery_urls.get(request_id)
+            if url:
+                click.echo(f"    Result file: {url}")
--- a/mech_client/cli/commands/request_cmd.py
+++ b/mech_client/cli/commands/request_cmd.py
@@ -29,6 +29,7 @@ from mech_client.cli.common import common_wallet_options, setup_wallet_command
 from mech_client.cli.validators import validate_chain_config, validate_ethereum_address
 from mech_client.services.marketplace_service import MarketplaceService
 from mech_client.utils.errors.handlers import handle_cli_errors
+from mech_client.utils.types import JSONValue
 from mech_client.utils.validators import (
     validate_batch_sizes_match,
     validate_extra_attributes,
@@ -36,7 +37,7 @@ from mech_client.utils.validators import (
 )
 
 
-def _dump(value: Any) -> str:
+def _dump(value: JSONValue) -> str:
     """
     Render a value parsed out of a delivered result file for CLI output.
 
@@ -46,7 +47,7 @@ def _dump(value: Any) -> str:
     return json.dumps(value, ensure_ascii=True, indent=2, sort_keys=True)
 
 
-def _format_delivery_output(delivery_data: Any) -> str:
+def _format_delivery_output(delivery_data: JSONValue) -> str:
     """
     Format delivery data for CLI output with parity across delivery modes.
 
--- a/mech_client/cli/commands/request_cmd.py
+++ b/mech_client/cli/commands/request_cmd.py
@@ -55,7 +55,7 @@ def _format_delivery_output(delivery_data: JSONValue) -> str:
     :return: The mech's answer, ready to print.
     """
     if delivery_data is None:
-        return "unavailable — could not read the result file (URL below)"
+        return "unavailable — could not read the result file"
 
     if isinstance(delivery_data, dict) and "result" in delivery_data:
         result = delivery_data["result"]
--- a/mech_client/domain/delivery/models.py
+++ b/mech_client/domain/delivery/models.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # ------------------------------------------------------------------------------
 #
-#   Copyright 2025 Valory AG
+#   Copyright 2026 Valory AG
 #
 #   Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
@@ -17,18 +17,25 @@
 #
 # ------------------------------------------------------------------------------
 
-"""IPFS infrastructure for uploading and downloading files via IPFS gateway."""
+"""Models shared by the delivery watchers."""
 
-from mech_client.infrastructure.ipfs.client import IPFSClient
-from mech_client.infrastructure.ipfs.metadata import push_metadata_to_ipfs
-from mech_client.infrastructure.ipfs.result_file import (
-    build_result_file_url,
-    fetch_result_file,
-)
+from dataclasses import dataclass
+from typing import Any, Optional
 
-__all__ = [
-    "IPFSClient",
-    "build_result_file_url",
-    "fetch_result_file",
-    "push_metadata_to_ipfs",
-]
+
+@dataclass(frozen=True)
+class DeliveryResult:
+    """A mech delivery, resolved to its content.
+
+    Both watchers return this, so callers get the same shape whether the
+    response was delivered on-chain or off-chain.
+
+    ``data`` holds the parsed content of the delivered result file, or
+    ``None`` when the gateway could not be read (``url`` still points at it).
+    ``url`` is the gateway URL ``data`` was read from; it is ``None`` for
+    offchain mechs that answer inline instead of pinning a result file.
+    """
+
+    request_id: str
+    data: Any = None
+    url: Optional[str] = None
--- a/mech_client/domain/delivery/offchain_watcher.py
+++ b/mech_client/domain/delivery/offchain_watcher.py
@@ -22,7 +22,7 @@
 import asyncio
 import logging
 import time
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 import requests
 from mech_client.domain.delivery.base import DeliveryWatcher
@@ -71,6 +71,11 @@ class OffchainDeliveryWatcher(
         :return: Dictionary mapping request ID to its delivery result
         """
         results: Dict[str, DeliveryResult] = {}
+        # Result-file URLs seen for requests whose file has not been read yet.
+        # A read that keeps failing is retried for as long as there is budget,
+        # so the URL is held here to report at timeout rather than written into
+        # `results`, which would mark the request done after a single attempt.
+        pending_urls: Dict[str, str] = {}
         prev_count = -1
         start_time = time.time()
 
@@ -93,15 +98,32 @@ class OffchainDeliveryWatcher(
 
                 try:
                     response = await self._fetch_offchain_data(request_id_int)
-                    if response:
-                        results[request_id] = await self._resolve_delivery(
-                            request_id, request_id_int, response
+                    if not response:
+                        continue
+
+                    url = self._result_file_url(response, request_id_int)
+                    if url is None:
+                        # The mech answered inline; the envelope is the answer.
+                        results[request_id] = DeliveryResult(
+                            request_id=request_id, data=response
                         )
-                        logger.info(
-                            f"Received offchain response for request {request_id_int}"
+                    else:
+                        pending_urls[request_id] = url
+                        # Off the event loop: a gateway round-trip here would
+                        # otherwise stall every other request's poll.
+                        results[request_id] = DeliveryResult(
+                            request_id=request_id,
+                            data=await asyncio.to_thread(
+                                fetch_result_file, url, request_id
+                            ),
+                            url=url,
                         )
+                    logger.info(
+                        f"Received offchain response for request {request_id_int}"
+                    )
                 except Exception as e:  # pylint: disable=broad-except
-                    # Log error but continue polling
+                    # Log error but continue polling. Leaving the request out of
+                    # `results` is what keeps it eligible for the next cycle.
                     logger.error(
                         f"Error fetching offchain data for {request_id_int}: {e}"
                     )
@@ -118,41 +140,36 @@ class OffchainDeliveryWatcher(
                     prev_count = current_count
                 await asyncio.sleep(WAIT_SLEEP)
 
+        # A request whose file never became readable still reports where to look,
+        # the same shape the on-chain path returns for an unreadable result.
+        for request_id, url in pending_urls.items():
+            results.setdefault(
+                request_id, DeliveryResult(request_id=request_id, data=None, url=url)
+            )
+
         return results
 
     @staticmethod
-    async def _resolve_delivery(
-        request_id: str, request_id_decimal: str, response: Any
-    ) -> DeliveryResult:
+    def _result_file_url(response: Any, request_id_decimal: str) -> Optional[str]:
         """
-        Resolve an offchain response to the result it points at.
+        Locate the result file an offchain response points at.
 
         The endpoint answers with an envelope carrying ``task_result``, the
         IPFS hash of the directory the mech filed the result under. Reading
-        that file here gives callers the same content the on-chain watcher
-        returns. Mechs that answer inline (no ``task_result``) are passed
-        through unchanged.
+        that file gives callers the same content the on-chain watcher returns.
 
-        :param request_id: Request ID in hex, as used to key the results
-        :param request_id_decimal: Same request ID in decimal, the name of the
-            result file inside the delivery directory
         :param response: Raw response from the offchain endpoint
-        :return: The resolved delivery result
+        :param request_id_decimal: Request ID in decimal, the name of the
+            result file inside the delivery directory
+        :return: URL of the result file, or ``None`` for a mech that answered
+            inline and pinned no file
         """
         task_result = (
             response.get("task_result") if isinstance(response, dict) else None
         )
         if not isinstance(task_result, str) or not task_result:
-            return DeliveryResult(request_id=request_id, data=response)
-
-        # Off the event loop: this is a gateway round-trip inside the polling
-        # loop, so blocking here would stall every other request's poll.
-        url = build_result_file_url(task_result, request_id_decimal)
-        return DeliveryResult(
-            request_id=request_id,
-            data=await asyncio.to_thread(fetch_result_file, url),
-            url=url,
-        )
+            return None
+        return build_result_file_url(task_result, request_id_decimal)
 
     async def _fetch_offchain_data(self, request_id: str) -> Any:
         """
--- a/mech_client/domain/delivery/offchain_watcher.py
+++ b/mech_client/domain/delivery/offchain_watcher.py
@@ -94,7 +94,7 @@ class OffchainDeliveryWatcher(
                 try:
                     response = await self._fetch_offchain_data(request_id_int)
                     if response:
-                        results[request_id] = self._resolve_delivery(
+                        results[request_id] = await self._resolve_delivery(
                             request_id, request_id_int, response
                         )
                         logger.info(
@@ -121,7 +121,7 @@ class OffchainDeliveryWatcher(
         return results
 
     @staticmethod
-    def _resolve_delivery(
+    async def _resolve_delivery(
         request_id: str, request_id_decimal: str, response: Any
     ) -> DeliveryResult:
         """
@@ -145,9 +145,13 @@ class OffchainDeliveryWatcher(
         if not isinstance(task_result, str) or not task_result:
             return DeliveryResult(request_id=request_id, data=response)
 
+        # Off the event loop: this is a gateway round-trip inside the polling
+        # loop, so blocking here would stall every other request's poll.
         url = build_result_file_url(task_result, request_id_decimal)
         return DeliveryResult(
-            request_id=request_id, data=fetch_result_file(url), url=url
+            request_id=request_id,
+            data=await asyncio.to_thread(fetch_result_file, url),
+            url=url,
         )
 
     async def _fetch_offchain_data(self, request_id: str) -> Any:
--- a/mech_client/domain/delivery/offchain_watcher.py
+++ b/mech_client/domain/delivery/offchain_watcher.py
@@ -75,6 +75,8 @@ class OffchainDeliveryWatcher(
         # A read that keeps failing is retried for as long as there is budget,
         # so the URL is held here to report at timeout rather than written into
         # `results`, which would mark the request done after a single attempt.
+        # Retrying matters most for the ordinary case — a pinned file that
+        # 404s until the gateway catches up — not just for unexpected errors.
         pending_urls: Dict[str, str] = {}
         prev_count = -1
         start_time = time.time()
@@ -111,12 +113,19 @@ class OffchainDeliveryWatcher(
                         pending_urls[request_id] = url
                         # Off the event loop: a gateway round-trip here would
                         # otherwise stall every other request's poll.
+                        data = await asyncio.to_thread(
+                            fetch_result_file, url, request_id
+                        )
+                        if data is None:
+                            # Not readable yet. `fetch_result_file` returns
+                            # `None` for every HTTP failure, and a freshly
+                            # pinned file routinely 404s while it propagates,
+                            # so leave the request out of `results` to have the
+                            # next cycle try again. The backfill below reports
+                            # it with this URL if it never becomes readable.
+                            continue
                         results[request_id] = DeliveryResult(
-                            request_id=request_id,
-                            data=await asyncio.to_thread(
-                                fetch_result_file, url, request_id
-                            ),
-                            url=url,
+                            request_id=request_id, data=data, url=url
                         )
                     logger.info(
                         f"Received offchain response for request {request_id_int}"
--- a/mech_client/domain/delivery/offchain_watcher.py
+++ b/mech_client/domain/delivery/offchain_watcher.py
@@ -38,6 +38,26 @@ logger = logging.getLogger(__name__)
 # Constants for offchain polling
 OFFCHAIN_DELIVER_ENDPOINT = "fetch_offchain_info"
 
+# `IPFS_URL_TEMPLATE` already carries the multibase and multicodec prefix, so a
+# delivery hash is the bare digest: 32 bytes, i.e. 64 hex characters.
+DELIVERY_HASH_BYTES = 32
+
+
+def _is_delivery_hash(value: str) -> bool:
+    """
+    Report whether a value is shaped like the hash of a delivery directory.
+
+    Checked by decoding rather than with ``int(value, 16)``, which would accept
+    a ``0x`` prefix the bare digest never carries.
+
+    :param value: Candidate ``task_result`` from the offchain endpoint
+    :return: True if it is a 32-byte digest in hex
+    """
+    try:
+        return len(bytes.fromhex(value)) == DELIVERY_HASH_BYTES
+    except ValueError:
+        return False
+
 
 class OffchainDeliveryWatcher(
     DeliveryWatcher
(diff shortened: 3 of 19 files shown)


mech_client/domain/delivery/onchain_watcher.py
```diff
@@ -69,28 +73,42 @@ def __init__(
...
             request_ids, request_id_to_mech, from_block
         )
 
+        # Step 3: Read the result files, concurrently and off the event loop.
+        # A batch delivers one file per request; reading them one after another
+        # would serialise that many gateway round-trips, each with its own
+        # timeout, after the caller's wait budget has already been spent.
+        file_data = await asyncio.gather(
+            *(asyncio.to_thread(fetch_result_file, url) for url in urls.values())
+        )
```
DIvyaNautiyal07, line 113: probe onchain gather 106

mech_client/services/marketplace_service.py
```diff
@@ -94,6 +98,27 @@ def _safe_int(value: Any, default: int = 0) -> int:
...
+    :return: Mapping with the ``delivery_results`` and ``delivery_urls`` keys.
+    """
+    return {
+        "delivery_results": {
+            request_id: result.data for request_id, result in results.items()
+        },
+        "delivery_urls": {
+            request_id: result.url for request_id, result in results.items()
+        },
+    }
```
DIvyaNautiyal07: probe marketplace 119

mech_client/domain/delivery/models.py
```diff
@@ -0,0 +1,41 @@
...
+
+    ``data`` holds the parsed content of the delivered result file, or
+    ``None`` when the gateway could not be read (``url`` still points at it).
+    ``url`` is the gateway URL ``data`` was read from; it is ``None`` for
+    offchain mechs that answer inline instead of pinning a result file.
+    """
+
+    request_id: str
+    data: Any = None
+    url: Optional[str] = None
```
DIvyaNautiyal07, line 43: probe models 41

mech_client/domain/delivery/onchain_watcher.py
```diff
@@ -69,28 +73,42 @@ def __init__(
...
         """
         # Step 1: Wait for marketplace delivery (get mech addresses)
         request_id_to_mech = await self._wait_for_marketplace_delivery(request_ids)
 
         if not request_id_to_mech:
             return {}
 
-        # Step 2: Get IPFS URLs from mech contracts
-        return await self._fetch_data_urls_from_mechs(
+        # Step 2: Get result file URLs from mech contracts
```
DIvyaNautiyal07: probe onchain step 95

mech_client/domain/delivery/onchain_watcher.py
DIvyaNautiyal07, line 113: **P1** — `asyncio.gather` defaults to `return_exceptions=False`, so the first `fetch_result_file` that raises cancels every sibling task and re-raises. `watch()` returns nothing and the caller — who has already spent the whole delivery-wait budget — loses every partial result the batch produced for one bad gateway read.

Fix:
```python
file_data = await asyncio.gather(
    *(asyncio.to_thread(fetch_result_file, url) for url in urls.values()),
    return_exceptions=True,
)
return {
    request_id: DeliveryResult(
        request_id=request_id,
        data=data if not isinstance(data, BaseException) else None,
        url=url,
    )
    for (request_id, url), data in zip(urls.items(), file_data)
}
```

Related to the next comment on `result_file.py` — the `response.text` fallback there can raise, and that's precisely the escape path this guard needs to survive.

mech_client/infrastructure/ipfs/result_file.py
```diff
@@ -0,0 +1,76 @@
...
+        response = requests.get(url, timeout=timeout)
+        response.raise_for_status()
+    except requests.exceptions.RequestException as e:
+        logger.warning("Could not fetch delivery result from %s: %s", url, e)
+        return None
+
+    try:
+        return response.json()
+    except ValueError:
+        return response.text
```
DIvyaNautiyal07, line 89: **P1** — `response.text` fallback is outside the `except requests.exceptions.RequestException` guard on line 71-73. `response.text` can itself raise — `LookupError` on an unknown declared encoding, other decoding errors on a badly-served binary pin — and that escape propagates through `asyncio.to_thread` into `gather` and (combined with the sibling comment on `onchain_watcher.py:106`) discards the whole batch.

Docstring on line 61-64 promises `None` on any read failure. Fix: broaden the inner catch so the docstring holds:

```python
try:
    return response.json()
except ValueError:
    pass
try:
    return response.text
except Exception as e:  # pylint: disable=broad-exception-caught
    logger.warning("Could not decode delivery result body from %s: %s", url, e)
    return None
```

mech_client/services/marketplace_service.py
DIvyaNautiyal07: **P2** — `_delivery_payload` splits a typed `Dict[str, DeliveryResult]` into two parallel flat dicts (`delivery_results`, `delivery_urls`) that the CLI at `request_cmd.py:254-263` immediately re-pairs by key. Parallel dicts must be kept in lockstep by convention, not by type.

Since this PR is already a documented breaking change to the API surface, this is the moment to expose the unified `Dict[str, DeliveryResult]` directly — one key, one object, `.data` and `.url` on the object. Removes `_delivery_payload`, the sync requirement, and the caller's zip-by-key pattern; the type system now enforces per-request pairing.

mech_client/domain/delivery/models.py
DIvyaNautiyal07, line 43: Nit — `DeliveryResult(request_id="x")` is structurally valid with both `data=None, url=None` defaults, but that state is semantically meaningless after a completed delivery (should be either "content unreadable, URL known" or "answered inline, no URL", never nothing). Optionally add:

```python
def __post_init__(self) -> None:
    assert self.data is not None or self.url is not None, \
        "DeliveryResult must have content or a URL"
```

Second nit: `data: Any` is very loose for what is a parsed JSON payload. `Optional[Union[Dict[str, Any], list, str]]` would bound it to JSON-compatible types and give mypy something to check at call sites.

mech_client/domain/delivery/onchain_watcher.py
DIvyaNautiyal07: Nit — the `# Step 1:` / `# Step 2:` labels above `_wait_for_marketplace_delivery` and `_fetch_data_urls_from_mechs` restate what the method name already says. The Step 3 comment earns its place (explains the concurrency rationale), but 1 and 2 are noise. Drop them.

mech_client/domain/delivery/offchain_watcher.py
```diff
@@ -112,6 +120,36 @@ async def watch(self, request_ids: List[str]) -> Dict[str, Any]:
...
+        :return: The resolved delivery result
+        """
+        task_result = (
+            response.get("task_result") if isinstance(response, dict) else None
+        )
+        if not isinstance(task_result, str) or not task_result:
+            return DeliveryResult(request_id=request_id, data=response)
+
+        url = build_result_file_url(task_result, request_id_decimal)
+        return DeliveryResult(
```
bennyjo: This `fetch_result_file` call blocks the event loop for up to 30s per delivered response inside the polling loop, so a batch of offchain deliveries serialises the same gateway round-trips the on-chain path's Step-3 comment avoids. Wrapping it in `asyncio.to_thread` (as the on-chain path does) would make the PR's "concurrent and off the event loop" claim true on both paths.

bennyjo (commented):
One note on the offchain path, complementing Divya's threads.

bennyjo (dismissed):
Round-2 changes verified: the off-chain result-file read now runs through asyncio.to_thread so the polling loop no longer blocks, gather degrades a bad read to data=None with the URL kept instead of sinking the batch, and the single deliveries map of DeliveryResult objects removes the parallel-dict drift risk. The remaining response.text guard question lives in Divya's open thread and its blast radius is now contained on both paths.

mech_client/domain/delivery/onchain_watcher.py
```diff
@@ -69,28 +73,50 @@ def __init__(
...
+        # serialise that many gateway round-trips, each with its own timeout,
+        # after the caller's wait budget has already been spent.
+        #
+        # `return_exceptions` keeps one bad read from sinking the batch: the
+        # requests are paid for by the time we get here, so a file that cannot
+        # be read degrades to `data=None` — its URL still points at it — rather
+        # than discarding every sibling result along with it.
+        file_data = await asyncio.gather(
+            *(asyncio.to_thread(fetch_result_file, url) for url in urls.values()),
+            return_exceptions=True,
```
DIvyaNautiyal07, line 112: **P1 — silent failure.** `return_exceptions=True` catches any `BaseException` and the comprehension at line 114 collapses it to `data=None` with no log at all. If `fetch_result_file` raises anything outside its own `try/except` (e.g. `AttributeError`, a threading exception), the paid delivery is silently degraded and the operator has zero diagnostic. Suggest logging at `error` with the request id, url, and exception repr before setting `data=None`.

mech_client/domain/delivery/offchain_watcher.py
```diff
@@ -88,7 +94,9 @@ async def watch(self, request_ids: List[str]) -> Dict[str, Any]:
                 try:
                     response = await self._fetch_offchain_data(request_id_int)
                     if response:
-                        results[request_id] = response
+                        results[request_id] = await self._resolve_delivery(
```
DIvyaNautiyal07: **P1 — asymmetric failure vs. onchain.** If this call raises (specifically the `asyncio.to_thread(fetch_result_file, url)` inside `_resolve_delivery`), the outer `except Exception` at line 103 logs it but never writes the key into `results`. The poll loop keeps retrying and the key eventually gets dropped from the returned dict entirely. The onchain path keeps the key with `data=None` and the url. The PR description explicitly claims parity between the two paths; this breaks it. Suggest catching inside `_resolve_delivery` and returning `DeliveryResult(request_id, data=None, url=url)` so the caller can still retry manually.

mech_client/services/marketplace_service.py
```diff
@@ -183,7 +183,12 @@ async def send_request(  # pylint: disable=too-many-arguments,too-many-locals
             with the shortfall and retry once (only applies to the offchain path)
         :param extra_attributes: Extra attributes for metadata
         :param timeout: Timeout for delivery watching
-        :return: Dictionary with request results
+        :return: Dictionary with ``tx_hash``, ``request_ids``, ``receipt``
+            (all ``None`` on the offchain path), and ``deliveries`` mapping
```
DIvyaNautiyal07: **P1 — docstring inaccuracy.** `request_ids` is not `None` on the offchain path — `_send_offchain_request` returns `"request_ids": request_ids_hex` at line 475. Only `tx_hash` and `receipt` are `None`. Suggest: `"...with tx_hash and receipt None on the offchain path..."`.

mech_client/infrastructure/ipfs/result_file.py
```diff
@@ -0,0 +1,76 @@
...
+    :param url: URL of the result file, as built by :func:`build_result_file_url`
+    :param timeout: HTTP timeout in seconds
+    :return: Parsed JSON content, the raw text if the file is not JSON, or
+        ``None`` if the gateway could not be read
+    """
+    try:
+        response = requests.get(url, timeout=timeout)
+        response.raise_for_status()
+    except requests.exceptions.RequestException as e:
+        logger.warning("Could not fetch delivery result from %s: %s", url, e)
```
DIvyaNautiyal07: **P2.** The warning logs only the url. The url does carry the decimal request id, but the rest of the codebase keys off hex — an operator investigating a failed delivery has to convert by hand. Consider having callers pass the hex request id in for logging.

mech_client/domain/delivery/models.py
```diff
@@ -0,0 +1,41 @@
...
+    response was delivered on-chain or off-chain.
+
+    ``data`` holds the parsed content of the delivered result file, or
+    ``None`` when the gateway could not be read (``url`` still points at it).
+    ``url`` is the gateway URL ``data`` was read from; it is ``None`` for
+    offchain mechs that answer inline instead of pinning a result file.
+    """
+
+    request_id: str
+    data: Any = None
```
DIvyaNautiyal07: **P2.** `data: Any` is broader than reality — every construction site passes a JSON-decoded value (dict, list, str, or `None` on failure). Tightening to `Optional[Union[dict, list, str]]` (or at least `Optional[Any]`) makes the type hint informative and turns misuse at consumers into a checker error instead of a silent runtime bug.

mech_client/services/marketplace_service.py
```diff
@@ -340,7 +345,7 @@ async def send_request(  # pylint: disable=too-many-arguments,too-many-locals
         return {
             "tx_hash": tx_hash,
             "request_ids": request_ids,
-            "delivery_results": results,
+            "deliveries": results,
```
DIvyaNautiyal07, line 401: **P2.** Return type is `Dict[str, Any]`, but the shape is now stable and documented as a public contract (this PR marks it as breaking). A `TypedDict` with the four keys (`tx_hash`, `request_ids`, `receipt`, `deliveries`) would make it checkable without changing behaviour.

mech_client/domain/delivery/models.py
```diff
@@ -0,0 +1,41 @@
...
+    Both watchers return this, so callers get the same shape whether the
+    response was delivered on-chain or off-chain.
+
+    ``data`` holds the parsed content of the delivered result file, or
+    ``None`` when the gateway could not be read (``url`` still points at it).
+    ``url`` is the gateway URL ``data`` was read from; it is ``None`` for
+    offchain mechs that answer inline instead of pinning a result file.
+    """
+
+    request_id: str
```
DIvyaNautiyal07, line 41: **Nit.** `request_id` on the dataclass duplicates the dict key it is stored under (every construction site does `results[request_id] = DeliveryResult(request_id=request_id, ...)`) and is never read anywhere in the diff. Dropping it removes a drift hazard with no loss of information.

mech_client/cli/commands/request_cmd.py
```diff
@@ -37,13 +36,40 @@
...
 
 
+def _dump(value: Any) -> str:
+    """
+    Render a value parsed out of a delivered result file for CLI output.
+
+    :param value: A value decoded from JSON, so always JSON-serialisable.
+    :return: Indented JSON.
+    """
+    return json.dumps(value, ensure_ascii=True, indent=2, sort_keys=True)
```
DIvyaNautiyal07, line 47: **Nit.** `_dump` is a one-line `json.dumps` with fixed kwargs and is called twice. Inlining `json.dumps(value, ensure_ascii=True, indent=2, sort_keys=True)` at both call sites reads fine and removes the indirection — worth a named helper when the kwargs might change or the call count grows.

mech_client/domain/delivery/onchain_watcher.py
```diff
@@ -69,28 +73,50 @@ def __init__(
...
+        )
+        return {
+            request_id: DeliveryResult(
+                request_id=request_id,
+                data=None if isinstance(data, BaseException) else data,
+                url=url,
+            )
+            for (request_id, url), data in zip(urls.items(), file_data)
```
DIvyaNautiyal07: **Nit.** `zip(urls.items(), file_data)` relies on `asyncio.gather` preserving input order (which it does) plus `urls` not being mutated between line 108 and here (it is not). Both hold today. Snapshotting once — `pairs = list(urls.items())` above the gather and iterating that — makes the ordering assumption explicit for the next reader.

mech_client/domain/delivery/offchain_watcher.py
```diff
@@ -112,6 +120,40 @@ async def watch(self, request_ids: List[str]) -> Dict[str, Any]:
...
+        request_id: str, request_id_decimal: str, response: Any
+    ) -> DeliveryResult:
+        """
+        Resolve an offchain response to the result it points at.
+
+        The endpoint answers with an envelope carrying ``task_result``, the
+        IPFS hash of the directory the mech filed the result under. Reading
+        that file here gives callers the same content the on-chain watcher
+        returns. Mechs that answer inline (no ``task_result``) are passed
+        through unchanged.
```
DIvyaNautiyal07: **Nit.** "Passed through unchanged" reads as if the raw envelope is returned directly. It is wrapped in `DeliveryResult(request_id=..., data=response)`. Suggest "wrapped as-is" or drop the phrase.

mech_client/domain/delivery/offchain_watcher.py
```diff
@@ -87,13 +98,32 @@ async def watch(self, request_ids: List[str]) -> Dict[str, Any]:
...
+                        results[request_id] = DeliveryResult(
+                            request_id=request_id, data=response
                         )
+                    else:
+                        pending_urls[request_id] = url
+                        # Off the event loop: a gateway round-trip here would
+                        # otherwise stall every other request's poll.
+                        results[request_id] = DeliveryResult(
+                            request_id=request_id,
+                            data=await asyncio.to_thread(
```
DIvyaNautiyal07: **Blocking — retry-on-unreadable-file is dead in production for HTTP failures.**

The comment at lines 74-77 documents intent: *"A read that keeps failing is retried for as long as there is budget, so the URL is held here to report at timeout rather than written into `results`, which would mark the request done after a single attempt."* But `fetch_result_file` catches `RequestException` and returns `None` (result_file.py:77-84). So on a real HTTP failure (404 during IPFS propagation, 503, timeout):

1. `fetch_result_file` returns `None` — no exception raised
2. This line runs: `results[request_id] = DeliveryResult(data=None, url=url)` — request marked done
3. Next poll iteration: line 96 `if request_id in results: continue` — skipped forever
4. The retry loop the whole `pending_urls` mechanism was built for never fires

The retry only triggers when `fetch_result_file` raises an *unexpected* exception (`UnicodeDecodeError`, `RuntimeError`, etc.). `test_raising_result_file_read_is_retried` uses `side_effect=RuntimeError` — a path production never takes. `test_unreadable_result_file_keeps_url` uses `return_value=None` and only asserts single-call behavior; retry is never verified.

**Fix (pick one):**
- Have `fetch_result_file` re-raise for a specific class of transient errors (5xx, `Timeout`, `ConnectionError`) so the outer `except Exception` catches them.
- Restructure this block so that when `data is None` the entry stays in `pending_urls` and is NOT written to `results`, keeping the request eligible for the next poll cycle.

Either way, add a test where `fetch_result_file.side_effect = [None, None, {"result": "ok"}]` and assert both `call_count > 1` and `results[rid].data == {"result": "ok"}` — that's the test that would have caught this.

mech_client/services/marketplace_service.py
```diff
@@ -94,6 +98,22 @@ def _safe_int(value: Any, default: int = 0) -> int:
...
+    """What :meth:`MarketplaceService.send_request` returns.
+
+    The shape is a documented public contract, so spell it out: a key typo at
+    a call site is then a mypy error rather than a runtime ``KeyError``.
+
+    ``tx_hash`` and ``receipt`` are ``None`` on the offchain path, which has no
+    transaction to report. ``request_ids`` is populated on both paths.
+    """
+
+    tx_hash: Optional[str]
```
DIvyaNautiyal07: **P2 — co-varying fields not enforced by the type.** `tx_hash` and `receipt` are always `None` together (offchain path, line 495/498) or always set together (on-chain path, line 367/370). The `TypedDict` allows `{"tx_hash": "0xabc", "receipt": None}` and `{"tx_hash": None, "receipt": {...}}` — both structurally valid but nonsensical.

The docstring documents the co-variance in prose, but prose isn't checked. A caller doing `if result["tx_hash"]:` still has to `assert result["receipt"] is not None` to satisfy mypy.

Consider a discriminated union:
```python
class OffchainRequestResult(TypedDict):
    tx_hash: None
    receipt: None
    request_ids: List[str]
    deliveries: Dict[str, DeliveryResult]

class OnchainRequestResult(TypedDict):
    tx_hash: str
    receipt: Dict[str, Any]
    request_ids: List[str]
    deliveries: Dict[str, DeliveryResult]

RequestResult = Union[OffchainRequestResult, OnchainRequestResult]
```
Lower priority since `Optional[str]` is functionally safe; upgrade only if you want mypy-enforced discrimination.

mech_client/cli/commands/request_cmd.py
```diff
@@ -27,23 +27,50 @@
...
+
+def _format_delivery_output(delivery_data: JSONValue) -> str:
+    """
+    Format delivery data for CLI output with parity across delivery modes.
+
+    :param delivery_data: Parsed content of the delivered result file.
+    :return: The mech's answer, ready to print.
+    """
+    if delivery_data is None:
+        return "unavailable — could not read the result file (URL below)"
```
DIvyaNautiyal07: **Nit — message promises a URL that may not appear.** The caller at line 262 only prints the `Result file:` line if `delivery.url` is truthy. If `data=None` and `url=None`, the user sees *"unavailable — could not read the result file (URL below)"* with nothing below.

This state isn't reachable today (all `data=None` construction sites set `url`), but the message is wrong by construction and will silently mislead the moment the reachability assumption breaks. Either drop the parenthetical:
```python
return "unavailable — could not read the result file"
```
or pass `url` into `_format_delivery_output` and make the phrase conditional on it.

tests/unit/domain/test_offchain_watcher.py
```diff
@@ -315,12 +316,176 @@ def mock_get_response(*args, **kwargs):
...
+        assert results["0x1a"].data == {"result": "the answer"}
+        assert mock_sleep.await_count >= 1
+
+    @pytest.mark.anyio
+    @patch("mech_client.domain.delivery.offchain_watcher.logger")
+    @patch("mech_client.domain.delivery.offchain_watcher.time.time")
+    @patch("mech_client.domain.delivery.offchain_watcher.asyncio.sleep")
+    @patch("mech_client.domain.delivery.offchain_watcher.fetch_result_file")
+    @patch("mech_client.domain.delivery.offchain_watcher.requests")
+    async def test_unreadable_result_file_reports_its_url_at_timeout(
```
DIvyaNautiyal07, line 435: **Nit — add a call-count guard so a regression fails meaningfully.** The `clock()` generator lets the poll loop enter `_resolve_delivery` once, then times out. If a future refactor moves the `fetch_result_file` call outside the `try:` block or otherwise short-circuits the read path, the test would `KeyError` on `results["0x1a"]` instead of failing with a clear message.

Add before the result assertions:
```python
assert mock_fetch_result_file.call_count >= 1
```
This pins that the polling loop actually reached the read path — so a regression fails at the right place with the right cause.

DIvyaNautiyal07 (changes requested):

mech_client/domain/delivery/onchain_watcher.py
```diff
@@ -69,28 +73,65 @@ def __init__(
...
+        )
+
+        results: Dict[str, DeliveryResult] = {}
+        for (request_id, url), data in zip(urls.items(), file_data):
+            if isinstance(data, BaseException):
+                # `fetch_result_file` logs the read failures it anticipates, so
+                # anything surfacing here is unforeseen. Say so loudly: the
+                # delivery is degraded either way, and a silent `None` would
+                # leave nothing to debug from.
+                logger.error(
```
DIvyaNautiyal07, line 125: **P1 — traceback lost.** `%r` renders the exception repr but not the stack, and `asyncio.gather(return_exceptions=True)` preserves `__traceback__` on the exception object — free to emit. This is the branch the comment calls out as "unforeseen — worth the noise"; the stack is the noise. Suggest:

```python
logger.error(
    "Unexpected error reading the result file for request %s from %s",
    request_id, url, exc_info=data,
)
```
Non-blocking, good to fix in a follow-up.

mech_client/domain/delivery/offchain_watcher.py
```diff
@@ -87,13 +100,39 @@ async def watch(self, request_ids: List[str]) -> Dict[str, Any]:
...
+                            request_id=request_id, data=data, url=url
+                        )
+                    logger.info(
+                        f"Received offchain response for request {request_id_int}"
+                    )
                 except Exception as e:  # pylint: disable=broad-except
-                    # Log error but continue polling
+                    # Log error but continue polling. Leaving the request out of
+                    # `results` is what keeps it eligible for the next cycle.
                     logger.error(
```
DIvyaNautiyal07: **P1 — traceback lost.** `f"...{e}"` gives just the message. With retries now handled higher up, anything reaching this handler is unforeseen and the stack is what you would want. Suggest `logger.exception(f"Error fetching offchain data for {request_id_int}")` (or the existing call with `exc_info=True`). Non-blocking, good to fix in a follow-up.

mech_client/domain/delivery/offchain_watcher.py
```diff
@@ -110,8 +149,37 @@ async def watch(self, request_ids: List[str]) -> Dict[str, Any]:
                     prev_count = current_count
                 await asyncio.sleep(WAIT_SLEEP)
 
+        # A request whose file never became readable still reports where to look,
+        # the same shape the on-chain path returns for an unreadable result.
+        for request_id, url in pending_urls.items():
+            results.setdefault(
```
DIvyaNautiyal07: **P2.** The backfill silently promotes unresolved requests to `data=None`. The timeout warning above only says `N/M received` — distinguishing "never delivered" from "delivered but file unreadable" then requires inspecting `.url` on each `DeliveryResult`. One log line per unresolved pending URL here would spare the caller that dance.

mech_client/domain/delivery/offchain_watcher.py
```diff
@@ -110,8 +149,37 @@ async def watch(self, request_ids: List[str]) -> Dict[str, Any]:
...
+            result file inside the delivery directory
+        :return: URL of the result file, or ``None`` for a mech that answered
+            inline and pinned no file
+        """
+        task_result = (
+            response.get("task_result") if isinstance(response, dict) else None
+        )
+        if not isinstance(task_result, str) or not task_result:
+            return None
+        return build_result_file_url(task_result, request_id_decimal)
```
DIvyaNautiyal07, line 218: **P2.** `_result_file_url` returns `None` only when `task_result` is missing/non-string/empty. A malfunctioning mech emitting e.g. `"pending"` or `"error"` as `task_result` yields a URL that 404s the full timeout budget. A hex/length sanity check (32-byte multihash → 64 hex chars) would let a non-hash string fall through to the inline branch instead of eating the timeout.

mech_client/services/marketplace_service.py
```diff
@@ -94,6 +108,39 @@ def _safe_int(value: Any, default: int = 0) -> int:
...
+class OnchainRequestResult(TypedDict):
+    """What :meth:`MarketplaceService.send_request` returns on the on-chain path.
+
+    The shape is a documented public contract, so spell it out: a key typo at
+    a call site is then a mypy error rather than a runtime ``KeyError``.
+    """
+
+    tx_hash: str
+    request_ids: List[str]
+    receipt: Dict[str, Any]
```
DIvyaNautiyal07: **P2.** Runtime value is `web3.types.TxReceipt` (an `AttributeDict`), not a plain `Dict[str, Any]`. The annotation is not a crash risk — `AttributeDict` is a dict subclass — but it misleads consumers about attribute-style access. Either type it as `TxReceipt` from `web3.types` or note the runtime shape in the docstring.

mech_client/domain/delivery/offchain_watcher.py
```diff
@@ -87,13 +100,39 @@ async def watch(self, request_ids: List[str]) -> Dict[str, Any]:
...
+                        continue
+
+                    url = self._result_file_url(response, request_id_int)
+                    if url is None:
+                        # The mech answered inline; the envelope is the answer.
+                        results[request_id] = DeliveryResult(
+                            request_id=request_id, data=response
                         )
+                    else:
+                        pending_urls[request_id] = url
```
DIvyaNautiyal07, line 133: **Nit.** `pending_urls[request_id]` is written before the fetch and never cleared on success. The backfill's `setdefault` covers it, but the invariant "`pending_urls` = unresolved only" would read more cleanly with a `pending_urls.pop(request_id, None)` after the successful write below.

tests/unit/domain/test_offchain_watcher.py
```diff
@@ -315,12 +316,185 @@ def mock_get_response(*args, **kwargs):
...
+        watcher = OffchainDeliveryWatcher(
+            mech_offchain_url="https://mech.example.com", timeout=60.0
+        )
+
+        results = await watcher.watch(["0x1a"])
+
+        assert mock_fetch_result_file.call_count == 2
+        assert results["0x1a"].data == {"result": "the answer"}
+        # ...and the unforeseen error was not silent
+        assert "gateway exploded" in str(mock_logger.error.call_args)
```
DIvyaNautiyal07: **Nit.** `"gateway exploded" in str(mock_logger.error.call_args)` is a substring check on the full `call(...)` repr, so it passes if the exception string shows up anywhere in the call. The onchain analog (`tests/unit/domain/test_delivery_watchers.py`) unpacks `call_args.args[N]` positionally — worth matching.

tests/unit/domain/test_offchain_watcher.py
```diff
@@ -315,12 +316,185 @@ def mock_get_response(*args, **kwargs):
...
+        # the case the retry exists for.
+        mock_fetch_result_file.side_effect = [None, None, {"result": "the answer"}]
+
+        watcher = OffchainDeliveryWatcher(
+            mech_offchain_url="https://mech.example.com", timeout=60.0
+        )
+
+        results = await watcher.watch(["0x1a"])
+
+        assert mock_fetch_result_file.call_count == 3
```
DIvyaNautiyal07, line 386: **Nit.** The test asserts `results["0x1a"].data` but not `.url`. A mutation that retries correctly but stops setting `.url` on the successful `DeliveryResult` would still pass. One extra assert covers the pairing.

tests/unit/domain/test_offchain_watcher.py
```diff
@@ -315,12 +316,185 @@ def mock_get_response(*args, **kwargs):
...
+        assert results["0x1a"].data == {"result": "the answer"}
+        assert mock_sleep.await_count >= 1
+
+    @pytest.mark.anyio
+    @patch("mech_client.domain.delivery.offchain_watcher.logger")
+    @patch("mech_client.domain.delivery.offchain_watcher.asyncio.sleep")
+    @patch("mech_client.domain.delivery.offchain_watcher.fetch_result_file")
+    @patch("mech_client.domain.delivery.offchain_watcher.requests")
+    async def test_raising_result_file_read_is_retried(
```
DIvyaNautiyal07, line 397: **P2 — coverage gap.** No test covers the mixed batch: 2 requests, one file readable immediately, the other keeps returning `None` until timeout. That's the interaction where `results.setdefault` vs `__setitem__` in the backfill matters and where the loop-exit condition composes with a partially populated `results`.

DIvyaNautiyal07 (dismissed):
Approving — the delta addresses the earlier concerns cleanly (single `deliveries` key, typed `RequestResult` union, retry-on-`None`, loud logging on unforeseen onchain errors, hex request ID in the gateway warning). A handful of non-blocking follow-ups below, worth picking up in a small chase-PR.

DIvyaNautiyal07 (approved):
All follow-ups from the previous round are in: `exc_info=data` on the onchain error log and `logger.exception` on the offchain one, `_is_delivery_hash` rejecting status strings before they become a URL, per-request warning at the backfill, `TxReceipt` on the TypedDict, `pending_urls.pop` on success (backfill can now assign directly), and a mixed-batch test. Nothing new to add.

atepem (approved):

bennyjo (approved):
Round-3 changes verified: an unreadable result file now stays pending and is retried each poll cycle within the budget (the common gateway-propagation 404 case), with the URL backfilled at timeout so the answer stays reachable by hand; unforeseen read errors are logged loudly on both paths instead of degrading silently; and the typed RequestResult union plus JSONValue alias pin the public contract. This also contains the remaining response.text concern in practice on both paths.

